PR #131, opened 2014-08-19
Remove rendering of other glass blocks' sides.

oitsjustjose:
ss 2014-08-19 at 09 21 12


Breadabix:
What version was this implemented? Im using Forestry 2.2.8.4 (MC 1.5.2) and im seeing the sides of glass.

PR #284, opened 2014-10-14
Lead shouldn't go into Hunter's Backpack

ghost:
Lead going into Hunter's Backpack:
1) makes no sense, as its not a mob drop
2) is incredibly annoying when you dismount and/or mount a horse and want to tie him up

I've already signed the CLA.


mezz:
This has finally been fixed by adding better configs. Sorry for the wait!
https://github.com/ForestryMC/ForestryMC/issues/694

PR #468, opened 2015-01-20
Fix Still wasting fluid at near full output.

Nirek-K:


mezz:
Thanks for the PR!
Ended up fixing this slightly differently, so that it gets a "tank full" error message too.
https://github.com/ForestryMC/ForestryMC/issues/473

PR #484, opened 2015-01-29
Add missing crated combs, fruit, nuts,vanilla saplings,wood,red sand and podzol

Nirek-K:


src/main/java/forestry/plugins/PluginCore.java
```diff
@@ -309,6 +309,18 @@ protected void registerCrates() {
 		crateRegistry.registerCrate(Blocks.sapling, "cratedSaplings");
 		crateRegistry.registerCrate(Items.potato, "cratedPotatoes");
 		crateRegistry.registerCrate(Items.carrot, "cratedCarrots");
+		crateRegistry.registerCrate(new ItemStack(Blocks.log,1,1), "cratedSpruceWood");
+		crateRegistry.registerCrate(new ItemStack(Blocks.log,1,2), "cratedBirchWood");
+		crateRegistry.registerCrate(new ItemStack(Blocks.log,1,3), "cratedJungleWood");
+		crateRegistry.registerCrate(Blocks.log2, "cratedAcaciaWood");
```
mezz: Thanks for these!
You should keep similar items together and use the "new ItemStack(block, 1, meta)" way for each, so that it's very clear we have meta 0, 1, 2, 3, etc all in a row.

mezz:
Thanks Nirek! Merged https://github.com/ForestryMC/ForestryMC/commit/8597d67416ff5ae8479bd396d606bbff78b6535f

Issue #223, opened 2014-09-24
Farming support for Pam's Harvest Mod, Botaina, Thaumcraft, and other mods.


CovertJaguar:
There are many new mods out there that add various kinds of crops and trees. We should expand our support where possible.

None of these have been tested for current support, or investigated as for how much integration we should provide. Discuss below. I'll be updating the main post with info/concerns as the discussion progresses.

One thing we don't want to do is break the balance of a mod by allowing easy automation of something that should be rare and hard to farm.

General topics:
- [ ] Test generic handling of mods on this list, some mods may already be covered.
- [ ] Right-Click Harvest: Investigate for increased generic crop support.
- [ ] Generic Sapling Support: Somehow wrap generic saplings in low maturation speed sapling wrapper block that would allow support for any sapling in the OreDict.

Mods to support, if I missed any please comment below:
- [x] [Pam's HarvestCraft](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/1274678): Flowers, crops, bushes, trees. Two types of crops, normal and regrowth. Fermenter and Squeezer integration? Add Forestry Fruit to Smoothy and Jelly recipes, etc...Some recipes already supported via OreDict. Author looking to add support, perhaps provide help? Plugins for Forestry provided support in the past, integrate from there? High Priority.
- [x] [GrowthCraft](http://growthcraft.wikispaces.com/): Grapes, Rice, Bamboo, Hops, etc... Manually Planted Farms are good for things like Grapes. Work with Author: @MatrexsVigil
- [x] [Witchery](https://sites.google.com/site/witcherymod/): Adds various Trees and Plants, used in Brews and Rituals. Discuss balance. Treents will spawn and kill you for cutting down Trees. Rumored Forestry support.
- [x] [Magical Crops](http://magicalcrops.weebly.com/): Adds crops that mature into mob drops.
- [x] [ExtraBiomesXL](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/1282126): Many varieties of trees. May already support Forestry.
- [x] [Biomes O' Plenty](http://biomesoplenty.wikia.com/wiki/Trees): Many new trees and flowers.
- [x] [Plant Mega Pack Mod](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-mods/1293061): Tons of new plants.
- [ ] [Twilight Forest](http://twilightforest.wikispaces.com/): Adds various Trees and Plants. Some with useful and magical properties.
[...]

PR #488, opened 2015-01-30
ExUtil support:

Nirek-K:
New Managed Farm: Ender (Tube:Ender Eye)
Places Endstone, plants Enderlily seeds


Vexatos:
I wouldn't want the ability to automate Ender Lilies.
If this gets merged, there should be a config option that is turned off by default.

Adaptivity:
@Vexatos :+1: I think there nothing bad in adding it BUT there are should be a config option.

mezz:
Thanks, merged! https://github.com/ForestryMC/ForestryMC/commit/870906188cd8672eab145938e5a7ef2652c21a7c

CovertJaguar:
Related to issue #223

PR #493, opened 2015-02-02
Slight Orchard logic fixes

Nirek-K:
Slightly widen search grid to include outlying fruit
Don't skip over trunk on first level


mezz:
Thanks Nirek! 
Tested with no adverse effects or performance problems.

PR #581, opened 2015-03-31
Plant Mega Pack Stage 3

Nirek-K:
Add most basic flowers to plantable and usable by bees. This should include almost all flowers that don't 'grow' like crops.


mezz:
Thanks! Merged: https://github.com/ForestryMC/ForestryMC/commit/70d6db8bf27c720161d866f99444884e070b7f55

PR #583, opened 2015-04-02
Add Spinach.

Nirek-K:
Add HC crops to the Orchard Tube. Allows diverse crop fields that won't end up all 1 type.


mezz:
Thanks Nirek! Merged https://github.com/ForestryMC/ForestryMC/commit/38ea1061d1cbe2d21a1c2c917f320abbf2dbc2a9

PR #686, opened 2015-05-26
Change fence recipes to match vanilla's style

ganymedes01:
I know forestry is not in 1.8 yet, but why not use it's recipe pattern for fences?

As seen here: http://minecraft.gamepedia.com/Fence
Fences in 1.8 are crafted with an specific type of wood and sticks in the middle, yielding 3 fences per recipe.


mezz:
Makes sense!
How about keeping both recipes for now, and remove the old one when we're in 1.8? I don't want to confuse people too badly.
Also please read our cla and [sign the cla-signatures document](https://github.com/ForestryMC/ForestryMC/blob/dev/cla-signatures.txt) in this PR.

ganymedes01:
Is this okay?

mezz:
Looks great, thanks for the PR!

Issue #396, opened 2014-12-21
Replace Apiarist's Pipe with a sorting block


mezz:
> The Apiarist Pipe should be replaced with a block that is more generic and compatible with more inventory management systems. It should also support Trees and Butterflies.

asiekierka:
If this is ever done, notify me beforehand - I would like to reinstantiate the Apiarist's Pipe as part of BuildCraftCompat if it is removed from here.

mezz:
The block would basically be a retexture that has support for other mods in addition to BC.

asiekierka:
That, or manually present PRs for updating to BuildCraft versions as they come. In general, I am going to go to lengths to ensure that the Apiarist's Pipe stays, even as an alternative to a sorting block.

mezz:
@CovertJaguar thoughts on moving the apiarist's pipe to BuildCraftCompat?
Seems like a fine idea to me. Forestry can implement the sorting block and BuildCraftCompat can keep the nice looking pipe up to date.

CovertJaguar:
I'm fine with that.

Nedelosk:
I think you have moved the apiarist pipe from forestry to bc compat.

Nedelosk:
Any idea how the block should work ?

mezz:
This block should work similarly to the Apiarist Pipe in BuildcraftCompat.
The GUI should be pretty much the same.
When an item is inserted, it should immediately output the item to a side according to the gui configuration. It should not accept items that it can't output somewhere.

Nedelosk:
@mezz What do you think about the block model ? 
![2018-01-08_04 43 11](https://user-images.githubusercontent.com/10259214/34658346-79977b1c-f42e-11e7-97c0-c0a75663e3f2.png)

mezz:
Looks very good! I think I would prefer colored band to be at the end before the connection (2 voxels over from where it is now), and make the pipe a bit wider. It will have a more "simple" look that way which may fit better with Forestry.

Nedelosk:
![2018-01-08_13 52 11](https://user-images.githubusercontent.com/10259214/34671488-265d1c7a-f47b-11e7-96ab-98755c613a6b.png)
?

mezz:
I think that looks great!

PR #687, opened 2015-05-28
Update to latest BuildCraft

ganymedes01:
Not sure if you'd want to do this now, but whenever you decide to, it's done :3

### DIFF
--- a/src/main/java/forestry/pipes/proxy/ProxyPipes.java
+++ b/src/main/java/forestry/pipes/proxy/ProxyPipes.java
@@ -20,7 +20,7 @@ import forestry.pipes.PipeItemsPropolis;
 import forestry.plugins.PluginPropolisPipe;
 
 import buildcraft.BuildCraftTransport;
-import buildcraft.core.CreativeTabBuildCraft;
+import buildcraft.core.BCCreativeTab;
 import buildcraft.transport.BlockGenericPipe;
 import buildcraft.transport.ItemPipe;
 import buildcraft.transport.Pipe;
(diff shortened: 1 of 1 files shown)


LouisHutch:
https://github.com/ForestryMC/ForestryMC/issues/396

mezz:
Not ready for this yet. If possible I'd like to move the pipe to BC Compat since it's much more tied to BC's updates than it is to Forestry's. Covert and Sengir are on board with the idea.

asiekierka:
@mezz: I think the move should be done in Forestry 3.6, personally, and the pipe kept in Forestry 3.5. That way users will get a grace period to move. Anyway, 3.5 still needs an update.

mezz:
I would like to cap the compatible BC Transport API version on Forestry so it does not crash on BC 7, and then add the pipe to BC Compat. It can handle the pipe for BC 7+.

asiekierka:
That's fine, but you should also add a missing mapping patch to one of these so worlds do not break.

mezz:
Yep, that's the first thing I got working :+1:

mezz:
PR to move Apiarist's pipe to BC Compat is complete: https://github.com/BuildCraft/BuildcraftCompat/pull/20
When it's pulled I will close this issue.

asiekierka:
Keep in mind to make the Apiarist Pipe module for Forestry 3.5.x stay for BC < 7.0.x. In 3.6, it should be removed.

mezz:
The Apiarist Pipe in Forestry is left active for BC 6 and disabled for BC 7 right now. If both mods have the pipe, there is no conflict. When their BC is upgraded to 7, all Apiarist Pipes become BC Compat.

Vexatos:
@asiekierka It should automatically remap if possible.

mezz:
It automatically remaps. If you get a forge warning that it's missing, be sure to update BC Compat before continuing.

PR #707, opened 2015-06-06
Altering Equivalent exchange intergration

border999:
It has bothered me for some time that even though ee2 was discontinued after 1.2.5, forestry still has not removed the integration and  updated it to the current ee3 which is still actively being updated on 1.7.10.
Also that notification that shows while the launcher is starting about not having ee2 is idiotic.
Anyone agree?


Vexatos:
....... what.

border999:
update the Equivalent exchange intergration

Parker8283:
I think he was looking at the fact that you have 100+ commits on this PR, which means you screwed up somewhere.

mezz:
EE2 integration has been removed on the latest dev already.
Also, there is already some work on EE3 integration here: https://github.com/ForestryMC/ForestryMC/commits/ee3_integration

Your PR is applied incorrectly, I see tons of past commits but no additions from you.
Please submit any PRs against the `ee3_integration` branch.
Please [add yourself to the CLA-signatures document](https://github.com/ForestryMC/ForestryMC/blob/dev/cla-signatures.txt) in your PR if you do.

Issue #806, opened 2015-08-24
Make chests use the same model as the vanilla chest


buttilda:
Forestry's chests still use the ancient full-block chest. Would be nice to see them updated to the vanilla model with the animated lid.

I looked into doing it myself briefly a while back and I couldn't find a way to do it without breaking existing worlds. I'm not very familiar with Forestry's source, so maybe it was just me being a derp.

I'd be willing to do this myself if I could figure out how to do it without breaking existing worlds (perhaps someone could point me in the right direction?), or if I was told that doing so was ok (the recent changes to the planks will already break old worlds, so maybe this is the right time to sneak this in?)

Vexatos:
Actually they look quite cool together with [Cyclopean Chests](https://github.com/purpleposeidon/CyclopeanChests) :wink:

mezz:
I'm trying to clean up world-breaking changes before the 4.0 release, see the issue here: https://github.com/ForestryMC/ForestryMC/issues/795

If you want, you can implement this in a branch and we can figure it out later.

buttilda:
super :3 I'll get started :D

PR #808, opened 2015-08-26
[WIP] Chests now render akin to the vanilla chests

buttilda:
This is not done yet!

-What should I do with the old textures? I haven't found a way to prevent them from being registered and causing those awful checkered textures from showing up (as breaking particles).

-Can anyone help me with the new textures? I'm horrible at texturing =P I made them look like the current ones but I can't find a way to fit the top face in without making it look weird.

-Bounding boxes can't be set on a per-block basis on BlockBase. Since lepidopterology and arboriculture chests are the only ones registered that problem could be worked around, but for apiculture changing the bounding box would affect other blocks (like the apiaries). How should I proceed?

-This is fully compatible with the current version of Forestry! I thought I'd have to break compatibility but turns out I didn't :D

-Other than that it all works! The lid animates perfectly and the sound plays as well.


buttilda:
Added a commit with a suggestion on how to solve the bounding box problem.

I'll squishy everything after this is all decided :3 Leaving it like this makes it easier to follow.

mezz:
Just noting: This PR is for issue #806 

You did the textures right: delete the old ones and make sure the new ones have a different name for texture pack compatibility.

The old icons are registered in `MachineDefinition.registerIcons`. Unfortunately, the best way may be to subclass MachineDefintion and override that method. That should fix the breaking particles as well.

I can help with the top texture, but I can't do it right now and I'm about to be gone for short vacation.

Doing bounding box with the machine definition looks like the right way.

Code looks good! Thanks ganymedes.

buttilda:
I'm glad everything is of accordance :)
I'll get started on fixing the rest.

buttilda:
Gave the textures a go
screen shot 2015-08-27 at 19 03 52

All I did was scale the pattern down... so anyone who can do better please do feel free!

buttilda:
Also managed to fix the broken textures by simply not setting the face numbers in the MachineDefinition. Deleted all of the old textures but one, which will be used for the particles (I found that using the full chest texture creates weird looking particles, but that might just be me being picky). Let me know if you like that solution. Committing it all now.

MysteriousAges:
Damn, this looks good!

buttilda:
Thanks :3
Do you wanna merge it? I can squash the commits if you think this is good enough.

mezz:
Just got back from vacation. I would like to take a stab at improving the textures before merging this. It looks great!

mezz:
Updated the textures so the glass area takes up the whole top instead of resizing it.

mezz:
Committed here: https://github.com/ForestryMC/ForestryMC/commit/5cd2350c9bcf4b634b25938c56d31356d7afb212

PR #810, opened 2015-08-27
Fix rainmaker crash

buttilda:
Bumped into this by accident.
Breaking the rainmaker causes a crash with a null pointer exception due to `FuelManager.rainSubstrate.get(null)` returning null


mezz:
Looks good, thanks!

Issue #809, opened 2015-08-26
Stained Glass = Redundant?


buttilda:
Sorry if this has already been discussed somewhere. I did a search on the repository and couldn't find anything.

Since vanilla added it's own coloured glass having one as well that looks very similar seems rather unnecessary... Why not remove them (making existing ones turn into the vanilla blocks) ?

For the picky players that will say they like the look of the Forestry ones better than the vanilla ones... there's always resource packs. The Forestry team could even publish one itself.

The crafting recipe using the Thermionic Fabricator could be maintained, but yielding (or course) the vanilla glass. I think it would be cool to have it as an alternative method of acquiring it that doesn't require sand.

labpaw:
I'm almost completely sure that I suggested (or someone else) to remove stained glass too but the answer was "no". There are a lot of mods that adding their own stained glass and their authors won't remove it. However, maybe Forestry devs decision have changed.

buttilda:
The only mod I can think of that also adds the feature is MFR, but their stained glass is not only for decoration, so I think it's "more okay", since theirs has an intricate and expensive recipes (though I still think it's somewhat repetitive)

leagris:
Compared to Forestry stained glass. Vanilla stained glass recipe are OP.
I'd suggest forestry to suppress vanilla recipes for glass and yes, remove its now duplicate stained glass implementation.

SandGrainOne:
When mods alters how a player obtain any resource used by other mods, it produce a lot of discontent. It is possible with a small unknown mod or plugin, but a mod with an existing presence should be very carefull with things like that. 

That being said I do whish more mods would do just that. Complexifying the crafting of building materials. Give the player more substantial reasons to automate the process.

buttilda:
Could simply make it a config option whether or not to disable the vanilla recipe for the glass.

If you ask me... I think that since it's just a building material, there's no point being so concerned over it, it wouldn't break the balance of the mod if they suddenly became easier to get.

mezz:
Agreed, there is no real purpose for Forestry's stained glass any more.
We should not change vanilla's recipe, it would be too unexpected. How about boosting the Forestry recipe to create more stained glass than the vanilla recipe?

mezz:
Here's the previous comments on stained glass: https://github.com/ForestryMC/ForestryMC/issues/357

buttilda:
I like the idea of making the Forestry recipe produce more glass.
[...]

PR #811, opened 2015-08-27
[WIP] Removed stained glass

buttilda:
This refers to issue #809

Things missing for the PR to be complete:

-Decide how much Stained Glass should be produced by the Thermionic Fabricator (at the moment I set the value to a random 4)

-Implement the mapping replacement event. I tried but just couldn't get it to work... anything I did just corrupted the world. I'm hoping someone can give me some insight on how to do it.


mezz:
Confirmed that mapping from modded to vanilla doesn't work.
We'll probably have to add a ticking tile entity that replaces itself with vanilla glass.
I'm going to sleep on it and then ask around a bit to be sure.

buttilda:
Okay :) Let me know what you decide.
What about the amount of glass made by the thermionic fabricator? Is 4 enough or should I change it?

mezz:
I haven't found a satisfactory way to migrate the old blocks to the new type. I think it's best to take advantage of this 4.0 release to simply break the blocks.

mezz:
Merged here https://github.com/ForestryMC/ForestryMC/commit/e9674795060a678531bf4f7b476594f0b3f2baea

PR #815, opened 2015-09-06
Fixes to OreDictionary support in squeezer recipes 

bdew:
Squeezer recipes in forestry at some point switched to use oredict. I'm currently registering all my custom honey drops as dropHoney which makes any custom squeezer recipes for them not work.

Moreover the Squeezer doesn't actually checks for OD matches when accepting items, making the check in recipes useless.

After some discussion in #forestry-dev the proposed solution (by vexatos) was to scan the registry twice - first without checking OD and if that fails, again with OD check enabled.

This patch changes the following:
- Implements dual registry scan when looking for recipe, first without OD then if that fails with OD
- Fixes item input to accept items that match recipes with OD
- Fixes the NEI handler to show those recipes correctly

I have tested it with Forestry recipes and custom honeydrops from gendustry and it seems i didn't break anything :P


mezz:
Looks good, this is the same way the other ore-dict aware things work. Thanks!

PR #818, opened 2015-09-09
Fixes for 3.6.x - BeekeepingLogic overwriting queens with old data + OreDict use in Squeezer recipes

### LINKED ISSUE #817: [3.6.x] BeekeepingLogic overwrites queen with old data
Happens if you swap a queen in any kind of bee housing without taking the old one out, on next bee tick the new queen will be overwritten.

This happens because the _queen_ field in BeekeepingLogic will only be unset (and then reloaded) if the queen itemstack is null or not a queen item. As long as it stays any kind of queen it will keep using the old data, and overwrite the item after the next bee tick.

Only occurs in 3.6, the new code in 4.0 detects that the itemStack was changed and reloads the data.

I'm not sure if you plan to make any more 3.6 bugfix releases but if you do i can make a fix PR.

(I also would really like to backport #815 to 3.6 if there will be another release)

bdew:
- Fixed #817 by adding the check for 4.0 for ItemStack changes
- Adapted the fix in #815 to 3.6.x

### DIFF
--- a/src/main/java/forestry/apiculture/BeekeepingLogic.java
+++ b/src/main/java/forestry/apiculture/BeekeepingLogic.java
@@ -46,6 +46,7 @@ public class BeekeepingLogic implements IBeekeepingLogic {
 	private int breedingTime;
 	private int queenWorkCycleThrottle;
 	private IEffectData effectData[] = new IEffectData[2];
+	private ItemStack queenStack;
 	private IBee queen;
 	private IIndividual pollen;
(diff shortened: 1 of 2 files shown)


mezz:
Looks good, thank you bdew!

mezz:
Just uploaded Forestry 3.6.5 to Curse
http://minecraft.curseforge.com/mc-mods/59751-forestry/files

PR #827, opened 2015-09-13
Add another mode that locks resupply on 1 item left in each slot

CallMeFoxie:
Also fixes the delay on restocking inventory
Also moves delay into NBT to avoid the up & down animation on changing dmg value

PS: I would love to sign your CLA, but all I get is "We're sorry, but something went wrong." upon entering my details ;)


CallMeFoxie:
Apparently it has still reordered some extra code even using your code format jar... :/

mezz:
I appreciate the work but I'm very reluctant to add any more modes to the backpack, it takes a while to switch between them all as-is. For new features, you should create an issue for discussion first.
You seem to have rearranged the order of the files significantly so it's more difficult to evaluate what changed. What is this new mode accomplishing?

If you would like to submit the animation change as a separate PR and sign our [cla-signatures.txt](https://github.com/ForestryMC/ForestryMC/blob/dev/cla-signatures.txt) I will accept it happily.

CallMeFoxie:
I used your code formatter settings, unsure why it reorganized the whole files as well ><

Basically it works just as resupply mode but leaves the last item in the slot, making sure that when you pick up items they won't go into the backpack into (now) emptied slots.

Another way would be to change the GUI and add a lock icon there... I can code that in instead, if that would make it better?

mezz:
Ah I see. Yes I think locking individual slots would be nicer, like the worktable memory slots.

When you reformat code there are options as checkboxes, one is for rearranging the code. We only care about the code formatting and the arrangement is whatever you want, so be sure to uncheck the code rearrangement box.

CallMeFoxie:
Kinda thought that it would be part of the formatting jar specifications, too lazy to check and uncheck checkboxes I fine-tuned for my devenv :) ... Will see what I can do then and rework it to some different way then... :) cheers

PR #835, opened 2015-09-19
Extracted Apiarist Armor checking methods to API so addons can use them more conveniently.

Vexatos:
Follow-up to ForestryMC/ForestryAPI#18.
This needs to be merged after ForestryMC/ForestryAPI#18.


Vexatos:
See #836.

PR #836, opened 2015-09-20
Extracted Apiarist Armor checking methods to API so addons can use them more conveniently.

Vexatos:
See #835 as well as ForestryMC/ForestryAPI#18.


mezz:
Looks good! 
I'm going to move the helper class to its own file but that's just my own preference.

mezz:
Merged here https://github.com/ForestryMC/ForestryMC/commit/614305c690b6802a64076b557f118bb704126b74

PR #863, opened 2015-10-03
FlowerRegistry and AgriCraft compatibility update

InfinityRaider:
I've correctly implemented the IPlantable interface into my BlockCrop class which allows to get the Block object corresponding with the planted seed on crop sticks as such: https://github.com/InfinityRaider/AgriCraft/blob/master/src/main/java/com/InfinityRaider/AgriCraft/blocks/BlockCrop.java#L723-L739

This allows for compatibility with the flower registry, regardless of which seed is planted.
I've kept the cases you guys added in for compatibility with older versions of AgriCraft.

### DIFF
--- a/cla-signatures.txt
+++ b/cla-signatures.txt
@@ -3,3 +3,4 @@ samtrion;samtrion@gmail.com;2015-01-18
 GenuineSounds;GenuineSoundsLtd@gmail.com;2015-02-26
 airbreather;airbreather@linux.com;2015-04-10
 ganymedes01;foka_12@hotmail.com;2015-05-27
+infinityraider;theoneandonlyraider@gmail;com;2015-10-03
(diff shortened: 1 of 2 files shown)


mezz:
Thanks! Looks good to me but I want to get @Nirek-K 's OK before we merge it in since he's the one who added the compatibility there.

mezz:
One last thing, can you agree to [the CLA](https://github.com/ForestryMC/ForestryMC/blob/dev/CLA.txt) in your PR? The file to add your signature is here:
https://github.com/ForestryMC/ForestryMC/blob/dev/cla-signatures.txt

InfinityRaider:
Sure, give me a second.

mezz:
Thanks! Squashed into 1 commit and merged: https://github.com/ForestryMC/ForestryMC/commit/8fb7d5eb672ee568116120de233ad310f750889d
It'll be available in the upcoming 4.0.6 release.

PR #971, opened 2015-11-18
Workaround to allow other mods to get plank icons without position

buttilda:
I have a mod called [WoodStuff](http://minecraft.curseforge.com/projects/wood-stuff) that adds several wood blocks (like bookshelves, chests, buttons, etc) made out of other mod's planks.
It assumes that planks have their icons determined using solely metadata (I thought it would be okay to do it this way because so far I hadn't found any mod that used TileEntities).

This simple workaround would be enough to fix Forestry's support for WoodStuff, so I hope that's okay.

### DIFF
--- a/src/main/java/forestry/arboriculture/blocks/BlockPlanks.java
+++ b/src/main/java/forestry/arboriculture/blocks/BlockPlanks.java
@@ -45,7 +45,10 @@ public class BlockPlanks extends BlockWood {
 	@SideOnly(Side.CLIENT)
 	@Override
 	public IIcon getIcon(int side, int meta) {
-		return IconProviderWood.getPlankIcon(EnumWoodType.LARCH);
+		EnumWoodType woodType = EnumWoodType.LARCH;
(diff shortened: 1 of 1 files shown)


buttilda:
Pretty sure I signed that already...
Edit: Done it again anyway :3

mezz:
Have you tried getting the icon from the Item?
`IIcon getIconIndex(ItemStack stack)` will return the correct icon.

buttilda:
I tried it and though it works for Forestry's planks, it doesn't work for other mods )=

This is what I currently do (in case you want to have a look):
https://github.com/ganymedes01/WoodStuff/blob/master/src/main/java/ganymedes01/woodstuff/blocks/BlockWoodPressurePlate.java

mezz:
Ok well as long as you took a look. I don't see any possible negative effects from this PR

buttilda:
Yeah, I can't think of anything this change would break.

Thank you for merging it :)

PR #1004, opened 2015-12-05
Add IC2 crops support to multifarms (uses orchard farm logic)

estebes:
I've added ic2 crop support to farms. I might work on an optimal solution in the future that includes a new farm logic to turn the farm into the crop harvester + crop-matron combo but for now this works prefectly.


mezz:
Thanks for the PR!
@Nirek-K can you take a look before I merge this?

Dream-Master:
@mezz is this in regular Forestry 4.2 for Minecraft 1.7.10 allready included?

liach:
Should be, according to [the build file](https://github.com/ForestryMC/ForestryMC/blob/1d924d5ed5b57050d038f540249790e72089dc1d/build.gradle#L14).

PR #1074, opened 2016-02-21
Fix Crate Models

Nedelosk:


mezz:
Thanks!
We need to remove the json requirement somehow, crates can be made with the API:
https://github.com/ForestryMC/ForestryMC/blob/mc-1.8.9/src/main/java/forestry/api/storage/ICrateRegistry.java

PR #1080, opened 2016-02-24
Fix Crate Models and other Models

Nedelosk:
and a test butterfly model as .obj and .b3d but that Model is at this moment not in the game.


mezz:
Thanks!

PR #1126, opened 2016-04-11
Mc 1.7.10 Fixed some null bug

xe5700:


mezz:
Thanks. 
Can you please match the style used in the rest of the mod? Full braces, the spacing of arguments and parenthesis, etc.

mezz:
@xjboss any update?

Issue #580, opened 2015-03-30
Add a way to register items to work as fertilizer in farms


Vexatos:
Currently Forestry's Apatite fertilizer is the only item that is allowed to be used as fertilizer in Forestry farms. It would be nice if there was a way to register other items as farm-compatible fertilizer; not sure if that would be good or if that would break the farms' balance too much...

mezz:
I think special care has been taken to prevent other fertilizers from working in the farm (no oredict name for fertilizer, no way to register fertilizers, few other uses for apatite, it's basically the only non-easily-renewable resource to power the farm). This seems to be against the intended design.

Adaptivity:
There are a lot of mods which adds own fertilizers. Not sure if it break balance since you shouldn't register *everything* that have fertilizer in name. Basically you should just add whitelist with acceptable fertilizers(which shouldn't break balance) and allow players to extend it to their taste.

mezz:
I see, that makes sense. Modpacks should have that flexibility.

Vexatos:
I was just thinking that there might be mods regisering their fertilizer as farm-compatible fertilizer when those types of fertilizer are much easier to get than the apatite one. Not sure if that would break balance; if it doesn't go ahead and add something to the API for that.

Yulife:
Yeah, API would be neat. Then Modtweaker could hook into that - custom fertilizers! Looking forward to it.

SirSengir:
The restriction to the non-renewable Apatite is intended design. Designs can change though. ;)

Problem with "fertilizer" added by mods usually is, that most of those are very, very easy to obtain and often extremly easy to make renewable - effectively invalidating the whole fertilizer mechanic. (Currently you need to find a way to make Apatite renewable and the only way I know of is a bee in MagicBees.)

If an API were added, it should offer the ability to adjust the effectiveness on a per-fertilizer basis. (I.e. have different fertilizers yield different fertilizerValues in https://github.com/ForestryMC/ForestryMC/blob/dev/src/main/java/forestry/farming/gadgets/TileFarmPlain.java#L650)

Vexatos:
@SirSengir That is exactly what I was worried about. Thanks for clarifying.

PR #1305, opened 2016-06-08
Add custom fertilizers support

estebes:
Implemented the ability to add custom fertilizers to farms. Also used this to add ic2 fertilizer support to farms as well as allowing it as fermenter fuel. Not sure if it is the best way of doing it but it probably can be used as a base for a better implementation.

### DIFF
--- a/src/main/java/forestry/Forestry.java
+++ b/src/main/java/forestry/Forestry.java
@@ -54,7 +54,7 @@ import forestry.plugins.PluginManager;
 				+ "after:BuildCraft|Energy@[7.2,);"
 				+ "after:ExtrabiomesXL;"
 				+ "after:BiomesOPlenty;"
-				+ "after:IC2@[2.3.235,);"
+				+ "after:IC2@[2.5.10,);"
 				+ "after:Natura@[2.2.0,);"
 				+ "after:HardcoreEnderExpansion;"
 				+ "after:JEI@[3.6.0,);"
--- a/src/main/java/forestry/api/farming/Fertilizers.java
+++ b/src/main/java/forestry/api/farming/Fertilizers.java
@@ -1,0 +1,19 @@
+package forestry.api.farming;
+
+import net.minecraft.item.Item;
+
+import javax.annotation.Nonnull;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author estebes
+ */
+public class Fertilizers {
+	/**
+	 * Used to add fertilizers to farms.
+	 * Params are an item (the fertilizer item) and the fertilizer value of said item.
+	 */
+	@Nonnull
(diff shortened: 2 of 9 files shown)


Vexatos:
If I remember correctly, I requested this long ago and it was rejected for some important reason. Not sure if it should become a thing (can't remember the exact reasoning).

Vexatos:
Found it: Issue https://github.com/ForestryMC/ForestryMC/issues/580.

Vexatos:
The main issue is that forestry fertilizer is supposed to be the only non-renewable thing you have to keep your farms supplied with (even if it's cheap and you get lots of fertilizer for little apatite). A major balance point (makes you still have to worry about _something_ after having set up a farm); most mods' fertilizers aren't built this way.

Nirek-K:
Eh, times change, as linked by that issue this was claimed as a milestone for 1.9 anyway. I already added something similar to 1.7 in a more limited fashion to not break apis and things.

Vexatos:
I still think it's something you should consider. I do appreciate it as a balance point, doesn't make the farms "set it and forget it".

SandGrainOne:
As long as there is a somewhat involving process that would need automation. Forestry is one of the best mods to include processes and cross system resource demands. It is something that should be cultivated.

Maybe also try to include new ways to use Apatite.

Forecaster:
It looks like this is hard coded?
Could you add a white-list system similar to the backpack configs for this?
I'd like to be able to keep it limited to apatite in my pack. I like auto-farming being tied to a finite resource.

And automation is still set-and-forget when it uses infinite resources as input.

Vexatos:
> And automation is still set-and-forget when it uses infinite resources like plants as input.

I fully agree with @Forecaster on this one. Fully automating a farm doesn't take long and is only mildly involved anyway.

CovertJaguar:
Yeah, user configurable (whitelist) is probably ok, but I'd really think twice about making it addon/mod configurable (IMC). There are a lot of mods out there that think it is funny to completely throw other mods balance out the window. And even more that wouldn't even give any thought to it beyond "oh! let's make our awesome new rock fertilizer compatible!", and not even look at the design behind it.

Vexatos:
Isn't IC2 fertilizer partially made of scrap? It would actually be cobblestone fertilizer :wink:

PrincessOfEvil:
Isn't Forestry fertilizer made of sand? I'm not sure what balance you're
talking about.

Vexatos:
@laz2727 Apatite.

Drullkus:
I'd like to have this. Still not merged, apparently.

Vexatos:
After the clarification in #580, I do not think this should be merged at all.

Nedelosk:
I close this, because we will not pull this.

PR #1373, opened 2016-08-02
Allow overriding climate when checking for mutations

bdew:
Before the big rewrite of mutations IBeeHousing could override the biome/temperature/humidity for mutation check purposes. This was used by Gendustry industrial apiaries with biome override upgrades.

After the rewrite the new MutationCondition\* classes get the biome directly from the world, not allowing for overriding from the apiary.

This PR changes that by adding a new interface IClimateProvider that lets mutation conditions check the climate while allowing it to be overriden. IBeeHousing and IButterflyNursery extend that interface while for trees a default implementation is provided in DefaultClimateProvider.

It includes some API changes, with the biggest one being IMutationCondition.getChance signature. The rest are probably not going to affect anybody.

I've discussed it on IRC with @Vexatos but didn't catch @mezz to talk about it. The relevant discussion is [here](https://gist.github.com/bdew/54bedfe5c1c6788b665f2f9817762f26).

I wasn't sure about exposing the default provider in the API, it could be useful for mods that have their own mutations (e.g. Botany). I could possibly add something like IClimateManager.getDefaultClimate(World world, BlockPos pos) if there's interest.

### DIFF
--- a/src/main/java/forestry/arboriculture/tiles/TileLeaves.java
+++ b/src/main/java/forestry/arboriculture/tiles/TileLeaves.java
@@ -507,7 +507,7 @@ public class TileLeaves extends TileTreeContainer implements IPollinatable, IFru
 
 	@Override
 	public EnumTemperature getTemperature() {
-		return EnumTemperature.getFromValue(ForestryAPI.climateManager.getTemperature(worldObj, pos));
+		return EnumTemperature.getFromBiome(worldObj.getBiome(pos), worldObj, pos);
 	}
 
 	@Override
--- a/src/main/java/forestry/arboriculture/tiles/TileLeaves.java
+++ b/src/main/java/forestry/arboriculture/tiles/TileLeaves.java
@@ -27,6 +27,9 @@ import forestry.api.arboriculture.ITree;
 import forestry.api.arboriculture.ITreeGenome;
 import forestry.api.arboriculture.ITreekeepingMode;
 import forestry.api.arboriculture.TreeManager;
+import forestry.api.core.EnumHumidity;
+import forestry.api.core.EnumTemperature;
+import forestry.api.core.ForestryAPI;
 import forestry.api.genetics.AlleleManager;
 import forestry.api.genetics.IAllele;
 import forestry.api.genetics.IEffectData;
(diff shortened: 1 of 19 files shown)


src/main/java/forestry/core/DefaultClimateProvider.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Override
+    public Biome getBiome() {
+        return world.getBiome(pos);
+    }
+
+    @Override
+    public EnumTemperature getTemperature() {
+        return EnumTemperature.getFromValue(ForestryAPI.climateManager.getTemperature(world, pos));
```
mezz: Replace this line with `EnumTemperature.getFromBiome(Biome Biome, World world, BlockPos pos)` here or it can't detect `EnumTemperature.HELLISH`

mezz:
Thanks for the PR! This is a good change.
`IClimateManager.getDefaultClimate` would be good, one less thing to worry about implementing.

I added a code comment for what I think is a bug. Please fix that and bump the `apiculture/package-info.java` version to 5.0.0 and I can pull this right away.

mezz:
Just an FYI, Binnie has retired and passed on the code to me. It's also visible source now. I don't have nearly enough time to fix it up for 1.10 but I will be cannibalizing parts of it for Forestry. I plan to make a simple bee addon that uses the bees from Extra bees, and add the woods to Forestry directly. At the pace I'm going it could take a while though, not a lot of free time recently.

Also, Myst is burned out on Magic Bees, probably since it's very hard to keep up with all the mod integration and will not be updating it to 1.10.

That basically leaves Forestry, Gendustry, and Computronics (Vexatos) as the main mods messing with bees in 1.10. Please feel very very free to discuss and break the API with any improvements you want while Forestry 5 is still in alpha.

bdew:
> I added a code comment for what I think is a bug. Please fix that and bump the apiculture/package-info.java version to 5.0.0 and I can pull this right away.

Done. Also added IClimateManager.getDefaultClimate.

mezz:
Cool thanks. BTW github added "squash and merge" button, so when you make a PR you don't have to worry about making it all 1 commit.

src/main/java/forestry/arboriculture/tiles/TileLeaves.java
```diff
@@ -495,4 +499,19 @@ public boolean canNurse(IButterfly butterfly) {
...
 	}
+
+	@Override
+	public Biome getBiome() {
+		return worldObj.getBiome(pos);
+	}
+
+	@Override
+	public EnumTemperature getTemperature() {
+		return EnumTemperature.getFromValue(ForestryAPI.climateManager.getTemperature(worldObj, pos));
```
mezz: Same problem as in DefaultClimateProvider 
Replace this line with `EnumTemperature.getFromBiome(Biome Biome, World world, BlockPos pos)` here or it can't detect `EnumTemperature.HELLISH`.

mezz:
Looks good, there's just one more issue with the Leaf Climate

bdew:
oops, fixed

PR #1407, opened 2016-08-23
Adding Greenhouse Window, Climate Control,  Camouflage Spray Can and Improve Recipes

Nedelosk:


Nedelosk:
@mezz  Is anything wrong with this PR ?

src/main/java/forestry/api/climate/IClimateSourceProvider.java
```diff
@@ -0,0 +1,19 @@
...
+package forestry.api.climate;
+
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+public interface IClimateSourceProvider {
+	
+	IClimateSource getClimateSource();
+	
+	World getWorld();
```
mezz: This name will cause issues.  You should have this interface extend `ILocatable`.

mezz:
I only see one thing, I made a comment.
This is more than I can review reliably. I can pull this if you are available to quickly fix any crashes that come up.

Nedelosk:
@mezz I have pulled this and I am available to quickly fix any crashes that come up.

PR #1449, opened 2016-10-07
Add model caching

asiekierka:
Or "lazy-loading", as I call it.


mezz:
Tested and looks like a big improvement, thanks asie!

Napkin:
Did adding this model caching also resolve the issues with supposedly missing textures of Forestry blocks when Optifine is used? I think it did! Definitely with release 267 all textures are visible. In 263 I couldn't make all textures show...

Thank you!

PR #1493, opened 2016-10-31
Crate Render Fix.

Speiger:
The Code is now much cleaner, Faster & Way More efficient.
If you want to copy textures do it like this. No reflection required. Also supports Color Multiplier properly. (Yes you can change easy the colormultiplier just by recreating the Quad without memory usage increase)
Also its caching all cases since 1 Quad is eating up like 116 bytes, And you will have only 3 of them * the the create amount. Still you will never get into critical issues with this system.

And if this crashes with a mod then Minecraft would also crash with the model they injected into Minecraft. So this is safe it "Should" never break in any case. "Note: Only code inside this class TRSRBakedModel not included"

Also thanks to my system of not recreating a quadList everytime you get asked for quads you save over time a lot of performance.
(Perfection OCD for the win)
Please import that as soon as possible. Because i really need this and i do not want to publish a custom forestry version.

### DIFF
--- a/src/main/java/forestry/core/models/ModelCrate.java
+++ b/src/main/java/forestry/core/models/ModelCrate.java
@@ -10,159 +10,59 @@
  ******************************************************************************/
 package forestry.core.models;
 
-import com.google.common.base.Charsets;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
-import org.apache.commons.io.IOUtils;
-
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.io.StringReader;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.IdentityHashMap;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
+import java.util.Map;
+
+import javax.annotation.Nullable;
 
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.block.model.BakedQuad;
 import net.minecraft.client.renderer.block.model.IBakedModel;
 import net.minecraft.client.renderer.block.model.ItemOverrideList;
-import net.minecraft.client.renderer.block.model.ModelBlock;
-import net.minecraft.client.resources.IResource;
-import net.minecraft.entity.EntityLivingBase;
-import net.minecraft.client.renderer.block.model.ModelBakery;
 import net.minecraft.client.renderer.block.model.ModelResourceLocation;
-import net.minecraft.client.renderer.vertex.DefaultVertexFormats;
-import net.minecraft.item.Item;
+import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.world.World;
-import net.minecraftforge.client.ItemModelMesherForge;
 import net.minecraftforge.client.event.ModelBakeEvent;
 import net.minecraftforge.client.model.IModel;
 import net.minecraftforge.client.model.IPerspectiveAwareModel;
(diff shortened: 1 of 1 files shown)


mezz:
Thanks for the contribution.
@Nedelosk does this look good to you?

src/main/java/forestry/core/models/ModelCrate.java
```diff
@@ -10,159 +10,59 @@
...
 
-	private static final Cache<String, IBakedModel> cache = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.MINUTES).build();
+	//Cache not needed because each baked quad is only 116 bytes big (minium size) and you do not have more then 1000 CrateTypes would only eat up 8 MB of ram.
+	private static Map<String, IBakedModel> cache = new HashMap<String, IBakedModel>();
 	private static final String CUSTOM_CRATES = "forestry:item/crates/";
 
-	private static IModel crateModel;
-	private static ModelBlock MODEL_GENERATED;
-    private static ModelBlock MODEL_ENTITY;
+//	private static IModel crateModel;
```
mezz: Please do not leave code commented out.

src/main/java/forestry/core/models/ModelCrate.java
```diff
@@ -10,159 +10,59 @@
...
-			}
-		} catch (IOException e) {
-			return false;
-		}
+//		try {
+//			crateModel = ModelLoaderRegistry.getModel(new ResourceLocation("forestry:item/crate-filled"));
+//		} catch (Exception e) {
+//			Log.error("Failed to init the crate model.", e);
+//		}
+//		ModelCrate.cache.invalidateAll();
```
mezz: Please do not leave code commented out.

mezz (changes requested):
Please do not leave code commented out.

Speiger:
Fixed.

Speiger:
@mezz is that ok?

Speiger:
just for the notes.
I did let it in just in case you need it for something.

src/main/java/forestry/core/models/ModelCrate.java
```diff
@@ -10,159 +10,51 @@
...
-import forestry.core.utils.Log;
 import forestry.storage.PluginStorage;
 
-import gnu.trove.map.hash.TIntObjectHashMap;
-
 @SideOnly(Side.CLIENT)
 public class ModelCrate extends BlankModel {
 
-	private static final Cache<String, IBakedModel> cache = CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.MINUTES).build();
+	//Cache not needed because each baked quad is only 116 bytes big (minium size) and you do not have more then 1000 CrateTypes would only eat up 8 MB of ram.
```
mezz: This note is not very clear in the code because it's on a field called "cache", and the comment seems to say it is not necessary. It's better to justify the reasoning here in the PR instead of in the code, unless you can find a better way to word it.

mezz (changes requested):

mezz:
Code can always be reverted, the history is in git, so there is no need to leave it there.

mezz:
Thanks, I thought Nedelosk might be online but I guess not. I will test this out.

Speiger:
@mezz its tested. I tested it with IC2Classic. No Crashes Etc. It does the same thing as before & It runs faster.

mezz:
I trust that you tested it, but I have to test it too. That's just how reviews work, the more people who can test the better.

mezz (approved):

mezz:
There are some visual issues when held and when in the world:

## Crates before (correct):





## Crates after (incorrect):

Speiger:
ok i take care of that too... I have already a idea how to solve that... The problem here is:
You need 2 models here...
1 the entity model & 1 Gui Model. Thats the problem... I take care of that tomorrow...

Speiger:
I fixed the item Held issue but the Entity is still scaled up trying too fix that... I hope i am able to

Speiger:
ok Fixed the entity version too. So i call this done...
Test it having fun with it and thanks for taking care of this^^

Speiger:
Also thanks to you i can now fix the size issue in classic too

mezz:
Looks good now, thanks.

PR #1641, opened 2017-03-25
Update RF_MOD_ID constant

disba1ancer:
Update RF_MOD_ID constant for compatibility with CoFHCore on MC 1.10.2


mezz:
Thanks.
Confirmed that the MOD_ID changed: https://github.com/CoFH/RedstoneFlux-API/blob/master/src/main/java/cofh/api/energy/package-info.java

PR #1673, opened 2017-04-26
Losslessly recompress PNG files using ECT

ibmibmibm:
* References
  - [compare](https://css-ig.net/png-test-corpus)
  - [homepage](https://github.com/fhanau/Efficient-Compression-Tool)


mezz:
Thanks!

What are the space savings of this PR? I already compressed png files using  PNGOUT, OptiPNG, and DeflOpt.

Do they show up correctly in-game? Some lossless png compressors result in grayscale-mode png files, which Minecraft handles incorrectly.

ibmibmibm:
I'll make a screenshot after my working hour.

ibmibmibm:
screenshot
Seems that item's textures are losing alpha channels, I'm checking what happened.
Other textures are worked, like blocks, machines, GUI.

ibmibmibm:
I've founded the problem: If I turn on the removal of colors from maximum alpha channel, the final image will result in a strange color of it. I'm going to regenerate this commit.

ibmibmibm:
It's a JDK bug:
https://bugs.openjdk.java.net/browse/JDK-6788458
I'm recompressing all files again without tRNS, and this should be solved.

mezz:
Thank you for checking.

leagris:
Note that further compression, optimizations of image assets bundled with a mod have few meaningful impacts:
- It certainly reduces size of the mod jar archive.
- It does not reduce runtime memory foot-print; because texture assets are loaded into uncompressed 32bit RGB pixmap buffers for the GPU to render.
- It can probably induce extra CPU cycles while loading these assets to: handle the extra compression, expand color indexes from palette table into 32bit RGB pixmap buffers.

Although, these optimizations are more valuable for high resolution resource packs distribution; they are not so for a mod jar archive.

ibmibmibm:
Before recompress:
4136799 forestry_1.11.2-5.3.3.0.jar
After recompress:
4084071 forestry_1.11.2-5.3.3.0.jar

leagris:
> Before recompress:
> 4136799 forestry_1.11.2-5.3.3.0.jar
> After recompress:
> 4084071 forestry_1.11.2-5.3.3.0.jar

- Saved 52728 bytes on the jar size
- No improvement of run-time memory foot-print.
- Added a bit more computations for resources loading.

@ibmibmibm As I said above: Your tools could be very beneficial for HD resource packs like example Soartex Moddded. Benefits are negligible for a mod jar.

mezz:
Part of the issue here is that I already compressed Forestry's textures using a slightly less powerful compressor, so space savings here are very low.
If you compress lots of big mods, it may make modpacks slightly faster to download. Some countries like Australia commonly have restrictive data caps as well. I agree the benefit isn't very big though.
Unless the computation cost is much higher, I don't see the harm if they want to spend the time to do this.

leagris:
So if you want to integrate png optimization, I suggest you do it through the gradle or other build process and keep 32Bit RGB assets as source. This automated optimization should intervene while packing assets into the final jar archive.
32Bit RGB image assets have the same value as source code. They are lossless and can be reprocessed as optimized/recompressed data for the packaged jar.

PR #1675, opened 2017-04-29
Fix names in BC|Transport compat

AlexIIL:
As with the update to 1.11 BC|Transport now has a modid of "buildcrafttransport", and the pipeWaterproof has been renamed to "waterproof".

You *might* want to rename this plugin from "BuildCraft 6 Transport" to "BuildCraft 8 Transport" - but that's minor.

The api for getting BC items is [here](https://github.com/BuildCraft/BuildCraftAPI/blob/8.0.x/api/buildcraft/api/BCItems.java#L123), so thet's where the registry names came from (and they won't be changed in 8.0.x or 7.99.x) in case I've made a mistake.

*EDIT:* I won't PR the other compat's until I'm certain that the BC|API won't change for those modules. I'm also not sure what todo with the refinery recipe as BC doesn't use a refinery any more for oil processing.


mezz:
Thanks!
I like to use `ForgeRegistries` instead of any mod's item API unless I have to, because it's less likely to crash from changes.
We'll probably rename the plugins to remove the version number completely, there's no real point to having it unless there are 2 major versions of BC for the same Minecraft version.

Issue #1743, opened 2017-07-13
Proposal: add "royal larvae" as a bee type


rmunn:
This is to support https://github.com/ForestryMC/Binnie/issues/200; the rationale behind this proposal can be explained by reading that issue, so I won't duplicate it here since it's rather long. I'm creating an issue over here because implementing https://github.com/ForestryMC/Binnie/issues/200 would require a new bee type in Forestry, so it would be good to have an issue in the Forestry repo in order to track that idea.

Also, if this works out well for 1.7.10, it might be nice to make the same change for 1.10 / 1.11 / 1.12. But I'm limiting the scope of this proposal to 1.7.10 at the moment since that's what @KorDum is currently working on for https://github.com/ForestryMC/Binnie/issues/200.

Nedelosk:
Same as https://github.com/ForestryMC/Binnie/issues/203 @rmunn @KorDum I am think this is a good idea. But not for 1.7.10. Forestry 1.7.10 is a final version and we only fixing major bugs of it and not adding any new features. We haven't changed anything on 1.7.10 since october 2016.
@mezz What do you thinking about adding this to 1.7.10 ?

rmunn:
I'd like to see it on 1.7.10 because that's what I'm still using (I can't give up Thaumcraft). If it gets added only to 1.10 or higher, that will still be a nice new feature for many people, but I won't be able to test it and give feedback.

mezz:
Sorry Forestry 1.7.10 is final, I'm reluctant to add new features even on the latest version.

rmunn:
#1744 (and https://github.com/ForestryMC/ForestryAPI/pull/24) looks like a pretty small patch, so I don't think it would create *too* much work to bring it in. It's just adding a single enum value. But you have more experience with Minecraft modding than I do: would adding a new enum value, that has never before been used, be likely to break mods all over the place? Or would it be relatively safe, and small?

mezz:
Binnie's is also not getting new features for 1.7.10 though.

rmunn:
The reason I ask, of course, is because having a royal larva bee type is needed for the change I proposed in https://github.com/ForestryMC/Binnie/issues/200. But if the "no changes in Forestry 1.7.10" decision is final, I'll have to come up with another way to get what I asked for in https://github.com/ForestryMC/Binnie/issues/200. Maybe the inoculator could be changed to take non-larval princesses as well as larval drones, but that breaks the mod's "feel".

[...]

PR #1744, opened 2017-07-13
Feature #1743

KorDum:
Feature #1743
Fixed redstone flux api submodule
Added Royal Larvae


rmunn:
Don't forget to update the Forestry API as well: the EnumBeeType definition lives in https://github.com/ForestryMC/ForestryAPI. Also, for minimal changes to other people's code that depends on the Forestry API, I strongly recommend adding the `EnumBeeType.ROYAL_LARVAE` definition **after** `EnumBeeType.NONE`.

If you put `EnumBeeType.ROYAL_LARVAE` *before* `EnumBeeType.NONE`, then it will have the value (in this case, 4) that `EnumBeeType.NONE` used to have. If that happens, any mod that was compiled against the old API and that compares things to `EnumBeeType.NONE` (which has the value 4 right now) will, when the player updates to the new Forestry version, suddenly see *actual bees* with an EnumBeeType value of 4 (because they're royal larvae). And that code may have something like this in it:

```java
if (beeType == EnumBeeType.NONE) {
    /* throw an exception because this is not supposed to happen */
}
```

And if other people's code has that in it, suddenly your royal larvae will start crashing people's games and breaking their worlds, because the value of `EnumBeeType.NONE` changed and made the other mod throw an exception.

However, if you put the `EnumBeeType.ROYAL_LARVAE` definition **after** `EnumBeeType.NONE`, then there's no problem. Other people's code may not know what to do with the number 5 showing up in EnumBeeType, but that's easy to update.

(This is all caused by the fact that Forestry's EnumBeeType has a classic beginner's mistake in it. If you have a `NONE` value in your enumeration, *it should always be the first value* so that its integer equivalent is 0. That way a value of 0 means "none", and more importantly, the integer value of `NONE` will never change! It's a classic newbie mistake to make `NONE` the *last* value in your enumeration, but then the integer value of `NONE` changes all the time if you have to add new values to your enum, and you have problems if you have to read old data. The lesson to be learned here is: the value of your `NONE` enum should *always* be 0, and the best way to do that is to make it the *first* value of your enum type, not the last. But it's too late to change Forestry's API now.)

KorDum:
@rmunn Oh, my mistake. I forget it.

PR #1746, opened 2017-07-14
Update modids for Natura and Extra Utilities 2

Obbi89:
It seems they follow the lowercase modid recommendation now.
It's a simple fix for newer versions.
This will break compatibility with older version of these mods.

Maybe it would be better to allow lower and uppercase names?


Nedelosk:
Thank you!

Obbi89:
I dun goofed while testing.
I guess I'll work on a fix for natura now.
There are some bugs in the initializiation (it is not working at all, cause either it never did or a lot changed)
When I'm on it I'll try to check if there are items missing etc. 
I'll open another pull request for that.

But first I want to try to add red orchid to the farm block.

Nedelosk:
I had to revert the pr, because it breaks the natura plugin.

Nedelosk:
@Obbi89 Can you please fix it and open a new one.

PR #1749, opened 2017-07-14
[1.10] Add Extra Utils Red Orchid support

Obbi89:


Nedelosk:
@mezz What do you think about the [recipe](https://github.com/ForestryMC/ForestryMC/pull/1749/files#diff-6bd78e5c3a5d4ab70de0ddb40f3362a8R241) ?

mezz:
The recipe looks very hard to me. Getting redstone ore is only possible after you have silk touch.
Maybe redstone blocks would be more appropriate.

PR #1750, opened 2017-07-16
Update to AA r115

Shadows-of-Fire:
The fluid "Oil" in actually additions was changed to "Refined Canola Oil" as of r115, to avoid conflicts with other mods using oil, as it is not equivalent.


mezz:
Thanks!

PR #1757, opened 2017-07-18
[1.10] Smaller fixes on IC², Natura, ExtraUtilities and BiomesOPlenty plugins

Obbi89:
~~Bop still has issues with Hummus.
They use their own detection for fertile soil.
Until then those Blocks will only work if we implement a internal Interface from bop.
This would make bop mandatory however. I don't see this as a solution.~~
A fix was merged upstream 👍 .
This patch is fixing the arbor.

This needs a upstream fix in bop.

Since it's all dependent on one another I closed the old PRs cause this combines: #1749 and #1755

### DIFF
--- a/src/main/java/forestry/plugins/compat/PluginNatura.java
+++ b/src/main/java/forestry/plugins/compat/PluginNatura.java
@@ -41,7 +41,7 @@ import forestry.plugins.ForestryPluginUids;
 @ForestryPlugin(pluginID = ForestryPluginUids.NATURA, name = "Natura", author = "SirSengir", url = Constants.URL, unlocalizedDescription = "for.plugin.natura.description")
 public class PluginNatura extends BlankForestryPlugin {
 
-	private static final String NATURA = "Natura";
+	public static final String modId = "Natura";
 
 	private static Block logNatura;
 	private static Block logWillow;
@@ -61,7 +61,7 @@ public class PluginNatura extends BlankForestryPlugin {
 
 	@Override
 	public boolean isAvailable() {
-		return ModUtil.isModLoaded(NATURA);
+		return ModUtil.isModLoaded(modId);
 	}
 
 	@Override
@@ -73,11 +73,11 @@ public class PluginNatura extends BlankForestryPlugin {
 	public void preInit() {
 		super.preInit();
 
-		logNatura = GameRegistry.findBlock(NATURA, "tree");
-		logWillow = GameRegistry.findBlock(NATURA, "willow");
-		leavesNatura = GameRegistry.findBlock(NATURA, "floraleaves");
-		saplingNatura = GameRegistry.findBlock(NATURA, "florasapling");
-		saplingNaturaRare = GameRegistry.findBlock(NATURA, "Rare Sapling");
+		logNatura = GameRegistry.findBlock(modId, "tree");
+		logWillow = GameRegistry.findBlock(modId, "willow");
+		leavesNatura = GameRegistry.findBlock(modId, "floraleaves");
+		saplingNatura = GameRegistry.findBlock(modId, "florasapling");
+		saplingNaturaRare = GameRegistry.findBlock(modId, "Rare Sapling");
 
 		ArrayList<String> saplingItemKeys = new ArrayList<>();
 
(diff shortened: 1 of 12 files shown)


src/main/java/forestry/plugins/compat/PluginNatura.java
```diff
@@ -72,45 +73,92 @@ public String getFailMessage() {
...
-		berryBlack = new ItemStack(berry, 1, 2);
-		berryMalo = new ItemStack(berry, 1, 3);
-		itemBarley = new ItemStack(barley);
+	private void consumeSubItems(Item item, String groupName, Collection<ItemStack> consumer) {
+		complexConsumeSubItems(item, groupName, subItem -> consumer.add(subItem));
+	}
+
+	private void complexConsumeSubItems(Item item, String groupName, Consumer<ItemStack> consumer) {
+		final ArrayList<ItemStack> subItems = new ArrayList<>();
+		item.getSubItems(item, CreativeTabs.SEARCH, subItems);
```
mezz: This method is client-only in 1.10 and will crash. It is only available on the server starting with 1.12.

src/main/java/forestry/plugins/compat/PluginNatura.java
```diff
@@ -119,145 +167,55 @@ public void registerCrates() {
...
+			cropBlocks.forEach(itemStack -> {
+				Block block = ItemStackUtil.getBlock(itemStack);
+				ItemStack seedItem;
+				int maxAge;
+				try {
+					// XXX: getMaxAge
+					maxAge = (int)block.getClass().getDeclaredMethod("getMaxAge").invoke(block);
+
+					// XXX: find seed items
+					seedItem = block.getPickBlock(block.getBlockState().getBaseState(), null, null, null, null);
```
mezz: This section is suuuper hacky.
I'd rather list out properties like (blockName, itemName, maxAge) for each crop instead of trying to get their properties automatically, unless there is a cleaner way.

mezz (changes requested):
Looks pretty good but there are some issues in the natura plugin to address before it can be merged.

src/main/java/forestry/plugins/compat/PluginNatura.java
Obbi89: Currently there is no clean way to bridge seed item to crop in Forge.
Which is a huge oversight in my opinion.

I tried to decouple the mod support for natura from actual hardcoded itemlists etc.
I guess that approach is not good enough for now. I'll see what or how I can change that to incorporate your suggestions.

src/main/java/forestry/plugins/compat/PluginNatura.java
Obbi89: Oh I didn't know that.
I guess I'll have to use run server in the future to find those errors.
Do you know if there is a way to see server/client-only methods in the IDE?

Nedelosk: The have a @SideOnly(Side.CLIENT) / @SideOnly(Side.SERVER) annotation.

Obbi89: Well this breaks everything. Since iterating over 2^16-1 possible subItems is just too much to be efficient. And as far as I see there is no other simple way to get sub items on the server side.
I would postpone this change to 1.12.

Since a lot of the code depends on that function, this is killing Natura support for 1.10.
Cause I will not add 60 hard coded items here.

Any suggestions?

Obbi89:
I removed the changes to the Natura plugin.
I guess I'll have to redo a special version just for 1.10 there :|

KnightMiner:
Couldn't you iterate through the seed items and use [IPlantable](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/common/IPlantable.java) to get the block? Thats what Tinkers Construct uses for crops and it works with Natura crops.

Obbi89:
I'll look into it later. Thanks.

src/main/java/forestry/plugins/compat/PluginNatura.java
mezz: [This plugin for intellij](https://plugins.jetbrains.com/plugin/8327-minecraft-development) will warn you about side-only issues. It has false-positives sometimes but I think it is useful.

As for fixing this, I think information has to be hard-coded. I assume they use meta values, so you can tell it a range to iterate over.

This will work fine in 1.12, so maybe you should target that version for now and come back to this later if you have more ideas.

mezz (approved):

PR #1780, opened 2017-07-29
Improve Greenhouse and fix Wood Pile

Nedelosk:


Nedelosk:
@mezz Can you please look at this if you have time ?

mezz:
This is much more than I can review correctly... I only have some recommendations.

I think you should move all the greenhouse climate stuff out of the `api`. I think it should be internal until you think it is complete and stable.

Make sure that the changes do not break old worlds.

Target 1.12, where packs are still pretty experimental and you will get good feedback.

PR #1794, opened 2017-08-11
Change event listeners to newer event system + sample fix for natura

Obbi89:
This is mainly for discussion on how to approach upcoming problems due to event api changes.

This is a trial and error solution to get a client up and running with changes introduced somewhere between 1.8 and 1.12.
Some mods (including Natura) are using RegistryEvents now. Those events are fired AFTER preInit making any preInit item look ups silently fail.
To fix this the inner workings of Forestry need to redesigned and moved around.
There are some unresolved issues (see comments)

I would use this as a discussion on how to tackle the problem, while providing a glance at some of the required changes.

This whole changeset is not really tested.
Especially everything model related etc. is totally untested. I do not have enough knowledge of the design of Forestry to see if moving some stuff around is actually the right thing to do here and there.

Additionally there will be some more changes to clean some mess up that is already in the code base. (See comments)

### DIFF
--- a/src/main/java/forestry/Forestry.java
+++ b/src/main/java/forestry/Forestry.java
@@ -15,7 +15,10 @@ import com.google.common.base.Preconditions;
 import javax.annotation.Nullable;
 import java.io.File;
 
+import forestry.plugins.compat.PluginNatura;
+import net.minecraftforge.client.event.ModelRegistryEvent;
 import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.event.RegistryEvent;
 import net.minecraftforge.fluids.FluidRegistry;
 
 import net.minecraftforge.fml.common.Mod;
@@ -26,6 +29,7 @@ import net.minecraftforge.fml.common.event.FMLInterModComms.IMCEvent;
 import net.minecraftforge.fml.common.event.FMLPostInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLPreInitializationEvent;
 import net.minecraftforge.fml.common.event.FMLServerStartingEvent;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
 import net.minecraftforge.fml.common.network.NetworkRegistry;
 import net.minecraftforge.fml.common.registry.GameRegistry;
 
@@ -58,7 +62,8 @@ import forestry.plugins.compat.PluginIC2;
 		acceptedMinecraftVersions = "[1.12]",
 		dependencies = "required-after:forge@[14.21.1.2395,);"
 				+ "after:jei@[4.7.0,);"
-				+ "after:" + PluginIC2.modId + ";")
+				+ "after:" + PluginIC2.modId + ";"
+				+ "after:" + PluginNatura.modId + ";")
 public class Forestry {
 
 	@SuppressWarnings("NullableProblems")
@@ -73,6 +78,7 @@ public class Forestry {
 		ForestryAPI.errorStateRegistry = new ErrorStateRegistry();
 		EnumErrorCode.init();
 		FluidRegistry.enableUniversalBucket();
+		MinecraftForge.EVENT_BUS.register(this);
 	}
 
 	@Nullable
@@ -105,6 +111,11 @@ public class Forestry {
 
 		PluginManager.runPreInit(event.getSide());
 
+	}
+
+	@SubscribeEvent
+	public void registerModels(ModelRegistryEvent event) {
+		PluginManager.runRegisterBackpacksAndCrates();
 		Proxies.render.registerModels();
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/forestry/plugins/BlankForestryPlugin.java
```diff
@@ -63,7 +63,20 @@ public void disabledSetupAPI() {
 	}
 
 	@Override
-	public void registerItemsAndBlocks() {
+	public void registerModels() {
+
```
Obbi89: TODO: fix whitespace error

src/main/java/forestry/plugins/PluginManager.java
```diff
@@ -210,12 +214,55 @@ public static void runSetup(FMLPreInitializationEvent event) {
...
+		for (IForestryPlugin plugin : loadedPlugins) {
+			Log.debug("Register Items Start: {}", plugin);
+			plugin.registerItems();
+			Log.debug("Register Items Complete: {}", plugin);
+		}
+
+		runRegisterCrates(event);
+	}
+
+	public static void runRegisterCrates(RegistryEvent.Register<Item> event) {
```
Obbi89: Todo use a functional approach like
```
run(name: string, consumer: Consumer<IForestryPlugin>)
  Log.debug(`{name] begin`)
  foreach plugin in loadedPlugins:
     consumer.apply(plugin)
     Log.debug(`    {plugin} ✓`)
  Log.debug(`{name} complete")
```

Obbi89: and then just run every stage with a lambda expression.
```run("registerItems", plugin => plugin.registerItems());```

File: src/main/java/forestry/plugins/PluginManager.java [hunk 3]
```diff
@@ -228,10 +275,6 @@ public static void runPreInit(Side side) {
 			if (ForestryAPI.enabledPlugins.contains(ForestryPluginUids.BUILDCRAFT_STATEMENTS)) {
 				plugin.registerTriggers();
 			}
-			if (ForestryAPI.enabledPlugins.contains(ForestryPluginUids.STORAGE)) {
-				plugin.registerBackpackItems();
-				plugin.registerCrates();
```
Obbi89: Those needed to be moved to the stage after registering blocks and items.
Maybe one should introduce new stages for this?

The Registry events are executed in alphabetical order by registry names except that items always run after blocks. 
(see forge api documentation)

src/main/java/forestry/Forestry.java
```diff
@@ -56,9 +60,10 @@
 		version = Constants.VERSION,
 		guiFactory = "forestry.core.config.ForestryGuiConfigFactory",
 		acceptedMinecraftVersions = "[1.12]",
-		dependencies = "required-after:forge@[14.21.1.2395,);"
+		dependencies = "required-after:forge@[14.21.1.2424,);"
```
mezz: Please pick the minimum version of Forge that will work, 2424 doesn't seem to have an important change. 
Also add a comment on this line saying which new feature of Forge is required.

src/main/java/forestry/plugins/compat/PluginNatura.java
```diff
@@ -25,93 +23,154 @@
 import forestry.core.fluids.Fluids;
 import forestry.core.recipes.RecipeUtil;
 import forestry.core.utils.ItemStackUtil;
+import forestry.core.utils.Log;
 import forestry.core.utils.ModUtil;
+import forestry.farming.FarmRegistry;
+import forestry.farming.logic.FarmableAgingCrop;
+//import forestry.farming.logic.FarmableSapling;
```
mezz: remove commented-out import

src/main/java/forestry/plugins/compat/PluginNatura.java
```diff
@@ -120,196 +179,74 @@ public void registerCrates() {
...
-		if (!berryMalo.isEmpty()) {
-			berries.add(berryMalo);
-		}
-
+			cropBlocks.forEach(itemStack -> {
+				Block block = ItemStackUtil.getBlock(itemStack);
+				ItemStack seedItem;
+				int maxAge;
+				try {
+					// XXX: getMaxAge
```
mezz: what does the `XXX` comment mean?

mezz (changes requested):
Looks good, needs some small changes or clarifications.

src/main/java/forestry/Forestry.java
Obbi89: I guess that one slipped in there... I need to check when or why I pushed the version. If not required I'll revert that change

src/main/java/forestry/plugins/compat/PluginNatura.java
Obbi89: XXX is the same as HACK.
Reflection is a hack in my eyes but regarding Minecraft modding in general I guess I can remove that note.

Maybe there is a better way to query custom attributes from items. But as far as I can see that's the easiest one.

Obbi89:
Ok changes should be applied

mezz (approved):

PR #1795, opened 2017-08-12
The End has no decoration step use old method there.

### LINKED ISSUE #1789: [1.12] Ender Hives no longer generate in the End
At the moment the hives in the end are not spawning anymore. I enabled the debug mode and while the overworld generates correctly, there are absolutely no ender hives in the End. Going back to version 5.4.3.137 generates the hives in the End, but the next version 5.4.3.144 doesn't, therefore it must be related to #1766.

Minecraft: 1.12
Forge: 14.21.1.244
Mods:
  forestry_1.12-5.5.0.154
  jei_1.12-4.7.5.84

Obbi89:
Fixes #1789


Obbi89:
Dirty :| but it seems the right way to not revert and break flat world again.

mezz:
Works for me, maybe we can make a cleaner solution later.

Obbi89:
revert and reapply :|

PR #1801, opened 2017-08-13
Reduce SimpleModel memory usage

asiekierka:
Immutable lists/sets have arrays which are already trimmed in size (as opposed to ArrayLists), as well as being much smaller with just a single entry in them.

This also fixes a small omission in SimpleModel.retexture.


Obbi89 (approved):

Obbi89:
I guess I should do this on a global scale later.

Nedelosk (approved):

asiekierka:
@Obbi89 Not necessarily. SimpleModel has a lot of instances in Forestry, but doing it in places with only a few instances might not be worth the effort.

PR #1804, opened 2017-08-14
MultiFarm fix for HarvestCraft plugin

serhiicherepanov:
In new implementation Items have name in lowercase, and forestry can't to find it.


mezz:
Thanks for the report!
Unfortunately it looks like this does not fix all of them.
I will take a look to see if I can match up everything.

mezz:
Fixed in the new builds for 1.10, 1.11, and 1.12.

Issue #1827, opened 2017-08-26
[1.12] Natura nether saplings can't be auto-farmed


LachimHeigrim:
Natura nether saplings - I tested Fuse Wood, Dark Wood and Ghost Wood - can't be autofarmed using multiblock farm. Tested manual planting and these saplings can't be planted on humus. They can be (manually) planted on dirt, netherrack or soul sand.

PR #1829, opened 2017-08-27
Fix for #1827: Natura nether saplings can be autofarmed in arboreal farm

LachimHeigrim:
Kindly review and comment. I'm new to Git, MC coding and GitHub so sorry if I messed up something unintentionally.

Summary of changes:
1. Added netherrack as valid soil for arboretum so that Natura nether trees can be autofarmed
2. Instead of hardocoding netherrack as soil for arboretum I wanted to add extensible mechanism so that plugins can add soils for farming logics. 
3. As farming logics are instantiated completely out of control of plugins, I opened `addSoil` to be public and allowed plugins to interact with farming logics after instantiaion through FarmRegistry.

Again, comments welcome.


mezz:
Thanks for contributing!
It looks like a lot of files have been edited without any real changes. 
I think your editor might have changed the line endings on those files (unix vs windows).
Can you fix that up so only the files you changed intentionally are submitted?

LachimHeigrim:
Thanks for catching it. Should be only the modified classes now.

mezz:
Looks good to me, thanks!

Issue #1847, opened 2017-09-10
Trade station does not check NBT when returning items


LunaticExperimentalist:
When generating the return mail from a trade the trade station appears to ignore NBT data when calculating items deducted. This can lead to forgery when working with currency and duplication when working with containers.

I have replicated this with versions 1.7.10-4.2.16.64 and 1.10.2-5.2.17.388

Process to replicate:
Make two stacks of currency. (e.g. emeralds) Use an anvil to name one stack "Real," and the other "Fake."
Setup a trade requiring Real as payment.
Fill a letter with a payment of Fake followed by Real, where the quantity of Fake is greater then or equal to the quantity of Real. The order of the currencies matters.
Complete the order.
The payee receives Real as payment, and the item offered is ducted from trade station.
The payer receives the item offered, the remainder of Fake, if any, and the Real offered for payment.

Issue #496, opened 2015-02-02
A few conveniences in the Trade Station and Catalog


marcin212:
A few conveniences in the Trade Station and Catalog:
- Ghost items or a search functionality (now you have to have the item you want to buy in the quantity you want to buy) Something like http://i.ytimg.com/vi/milg8tDclI8/maxresdefault.jpg
- ability to pull the received items out (pipes, hoppers, import busses, etc.)
- Ore Dictionary, NBT tags support, like http://teamcofh.com/wp-content/uploads/2014/07/GuiItemduct-300x187.png
- Search functionality in the Catalog

mezz:
1) Good idea, ghost items are a possibility. A search function is a bit too much though.

2) You can extract received items, but it has to be set to "public" since there is no way to know who owns the automation pipes.

3) I'm reluctant to make the trade station GUI more complicated but it seems like a good idea too.

4) What would you want to search for? Trader name, item offered, item requested, player name...?

marcin212:
4) maybe like NEI(@p <nickname>, @m <modname>)

PR #1850, opened 2017-09-15
Have trade station compare tags when accepting payment

LunaticExperimentalist:
Potential fix to #1847.


LunaticExperimentalist:
I'm going to comment that the change a made here "just works," but is maybe not optimal. The patch changes how Forestry compares ItemStacks at a low level that I thought logical. This patch will also effect how recipes are applied, and I don't know enough about the internals of Forestry to know whether that will cause more problems than it solves.
Another option here may be to give the trade station it's own code for comparing ItemStacks, and/or implement note 3 of #496.

mezz:
Thanks!
Unfortunately you are probably right in thinking that this could have unintended side effects, I'm not sure it's the right fix.

LunaticExperimentalist:
I made some improvements. There's a boolean flag to indicate that strict tag matching should be used. This should have zero effect on existing references to containsSets.

mezz:
Thank you, this looks much safer. This way we can look into the other machines another time.

PR #1892, opened 2017-11-12
Improve Genetic api

Nedelosk:
- Add Geneome Save Formats
   - UID: Used before a player joined a world and for backward compatibility
   - BINARY: Used after a player joined a world
- Fix Butterfly Item Model Scale
- Add `IndividualHandler` capability
- Add `AlleleTemplate` and `AlleleTemplateBuilder`


src/main/java/forestry/api/genetics/IAlleleRegistry.java
```diff
@@ -144,6 +146,13 @@
 	 */
 	Collection<IChromosomeType> getChromosomeTypes(IAllele allele);
 
+	/* SAVE AND LOAD*/
+
+	int getAlleleID(IAllele allele);
```
mezz: Is this needed in the API? 
I think dealing with `int` is dangerous, we should only expose it if we have to.
It may be better leave the saving and loading for Forestry to do internally.

src/main/java/forestry/api/genetics/IAlleleTemplate.java
```diff
@@ -0,0 +1,38 @@
...
+
+	/**
+	 * @return The size of the allele array.
+	 */
+	int size();
+
+	/**
+	 * @return A copy of this template with a copied allele array.
+	 */
+	IAlleleTemplate copy();
```
mezz: Should be `IAlleleTemplate<T, S>`?

src/main/java/forestry/api/genetics/IAlleleTemplate.java
```diff
@@ -0,0 +1,38 @@
...
+
+	ISpeciesRoot getRoot();
+
+	IIndividual toIndividual(@Nullable IAlleleTemplate inactiveTemplate);
```
mezz: `IAlleleTemplate<T, S>`

src/main/java/forestry/api/genetics/IAlleleTemplate.java
```diff
@@ -0,0 +1,38 @@
...
+
+	IGenome toGenome(@Nullable IAlleleTemplate inactiveTemplate);
```
mezz: `IAlleleTemplate<T, S>`

src/main/java/forestry/api/genetics/IAlleleTemplate.java
```diff
@@ -0,0 +1,38 @@
...
+
+	IChromosome[] toChromosomes(@Nullable IAlleleTemplate inactiveTemplate);
```
mezz: `IAlleleTemplate<T, S>`

src/main/java/forestry/core/genetics/Chromosome.java
```diff
@@ -22,22 +22,44 @@
...
+	public static Chromosome create(@Nullable String primarySpeciesUid, @Nullable String secondarySpeciesUid, IChromosomeType chromosomeType, @Nullable IAllele primary, @Nullable IAllele secondary) {
 		primary = validateAllele(primarySpeciesUid, chromosomeType, primary);
 		secondary = validateAllele(secondarySpeciesUid, chromosomeType, secondary);
 
 		return new Chromosome(primary, secondary);
 	}
 
+	@Nullable
+	public static IAllele getFirstAllele(NBTTagCompound chromosomeNBT){
+		String speciesUid = chromosomeNBT.getString(Chromosome.UID0_TAG);
```
mezz: `speciesUid` should be `alleleUid`

src/main/java/forestry/core/genetics/Chromosome.java
```diff
@@ -22,22 +22,44 @@
...
+		return AlleleManager.alleleRegistry.getAllele(speciesUid);
+	}
+
+	@Nullable
+	public static IAllele getSecondAllele(NBTTagCompound chromosomeNBT){
+		String speciesUid = chromosomeNBT.getString(Chromosome.UID0_TAG);
```
mezz: `speciesUid` should be `alleleUid`

src/main/java/forestry/core/genetics/GenomeSaveHandler.java
```diff
@@ -0,0 +1,82 @@
...
+		return writeFormat.writeTag(genome, tagCompound);
+	}
+
+	public static IChromosome[] readTag(NBTTagCompound tagCompound, ISpeciesRoot speciesRoot) {
+		SaveFormat format = getFormat(tagCompound);
+		return format.readTag(tagCompound, speciesRoot);
+	}
+
+	@Nullable
+	public static IChromosome getChromosomeDirectly(NBTTagCompound genomeNBT, IChromosomeType chromosomeType) {
```
mezz: We need `getAlleleDirectly(NBTTagCompound genomeNBT, IChromosomeType chromosomeType, boolean active)` instead. 
Ideally we should only read 1 nbt tag to find the species for rendering. Right now your `Genome.getAlleleDirectly` is loading a chromosome which requires reading 2 tags, one for each allele.

src/main/java/forestry/core/genetics/SaveFormat.java
```diff
@@ -0,0 +1,142 @@
...
+
+		@Override
+		public boolean canLoad(NBTTagCompound tagCompound) {
+			return tagCompound.hasKey("Chromosomes");
+		}
+	},
+	BINARY {
+		private static final String DATA_TAG = "data";
+		private static final String VERSION_TAG = "version";
+		private static final String VERSION = "1.0.0";
```
mezz: version would be smaller as an int

src/main/java/forestry/api/genetics/ISpeciesRoot.java
```diff
@@ -69,14 +69,35 @@
...
-	<O extends Object, I extends IIndividual> I translateMember(O objectToTranslator);
+	<I extends IIndividual> I translateMember(Object objectToTranslator);
+
+	/**
+	 * Translates {@link net.minecraft.block.state.IBlockState}s and {@link net.minecraft.item.ItemStack}s into genetic
+	 * data.
+	 *
+	 * @since Forestry 5.8
+	 */
+	ItemStack getGeneticEquivalent(Object objectToTranslator);
```
mezz: I don't really like that these work on `Object`, can you make them work on `ItemStack` and `IBlockState` parameters as separate functions? We will not have any other types so I don't think it needs to be very generic.

src/main/java/forestry/core/genetics/alleles/AlleleTemplateBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+		return this;
+	}
+
+	@Override
+	public IAlleleTemplateBuilder<T, S> set(T chromosomeType, IAllele allele){
+		AlleleHelper.getInstance().set(alleles, chromosomeType, allele);
+		return this;
+	}
+
+	public AlleleTemplateBuilder<T, S> set(T chromosomeType, Object value) {
```
mezz: The `value` parameter should probably be `IAlleleValue` type

src/main/java/forestry/core/genetics/GenomeSaveHandler.java
```diff
@@ -0,0 +1,82 @@
...
+
+	public static IChromosome getSpecificChromosome(NBTTagCompound genomeNBT, IChromosomeType chromosomeType) {
+		SaveFormat format = getFormat(genomeNBT);
+		return format.getSpecificChromosome(genomeNBT, chromosomeType);
+	}
+
+	/**
+	 * Converts the nbt into the current format.
+	 */
+	public static ItemStack covert(ItemStack itemStack){
```
mezz: I don't think we need this. Instead, we could have `GenomeSaveHandler.writeTag` always write using the latest format.

src/main/java/forestry/core/utils/SimpleByteBuf.java
```diff
@@ -0,0 +1,149 @@
...
+		IChromosomeType[] chromosomeTypes = speciesRoot.getKaryotype();
+		int types = readVarInt();
+		IChromosome[] chromosomes = new IChromosome[chromosomeTypes.length];
+
+		String primarySpeciesUid = null;
+		String secondarySpeciesUid = null;
+
+		for (IChromosomeType type : chromosomeTypes) {
+			int index = type.ordinal();
+			int bit = (1 << chromosomeTypes.length - index - 1);
```
mezz: What is this doing? I do not understand it.

mezz (changes requested):
This looks great! I have some request for changes, but this is very nice work.

src/main/java/forestry/core/genetics/GenomeSaveHandler.java
Nedelosk: Do we not need the whole chromosome to find out which of the two alleles is the active / dominant one ?

mezz: We should store the active alleles as the first in each pair.

Nedelosk: That means that the order of the two alleles doesn't matter anymore and we can check at the creation of the chromosome which of the two alleles is the active one?

mezz: I'm not sure the order matters. For picking which one is used we choose the active one of the pair, and for breeding we pick one randomly. I can't think of a real relation between the alleles in a chromosome, besides that they came from one parent. I think it should be ok to mix them up unless you see a reason not to.

src/main/java/forestry/api/genetics/IChromosome.java
```diff
@@ -5,21 +5,60 @@
...
+	IAllele getActiveAllele();
+
+	/**
+	 * @return The inactive allele of this chromosome.
+	 */
 	IAllele getInactiveAllele();
 
-	IAllele getActiveAllele();
+	/**
+	 * Crates a new chromosome out of the alleles of this chromosome and the other chromosome.
```
mezz: `Crates` should be `Creates`

src/main/java/forestry/api/genetics/IChromosome.java
```diff
@@ -5,21 +5,60 @@
...
+	 *
+	 * It always uses one allele from this and one from the other chromosome to create the new chromosome.
+	 * It uses {@link Random#nextBoolean()} to decide which of the two alleles of one chromosome it should use.
+	 *
+	 * @param rand 				 The instance of random it should uses to figure out which of the two alleles if should
```
mezz: This param description should be aligned with spaces

src/main/java/forestry/api/genetics/ISpeciesRoot.java
```diff
@@ -162,10 +162,24 @@
 
 	IGenome templateAsGenome(IAllele[] templateActive, IAllele[] templateInactive);
 
-	/* TEMPLATES */
+	/**
+	 * @return A wrapped version of the genome.
+	 *
+	 * @since Forestry 5.8
+	 */
+	default IGenomeWrapper getWrapper(IGenome genome){
```
mezz: should be `@Nullable`

src/main/java/forestry/core/genetics/Chromosome.java
```diff
@@ -22,140 +22,161 @@
...
+			if (template == null) {
+				template = speciesRoot.getDefaultTemplate();
+			}
+
+			return template[chromosomeType.ordinal()];
+		}
+		return allele;
+	}
+
+	private static IAllele getActiveAllele(IAllele firstAllele, IAllele secondAllele) {
```
mezz, line 160: This strategy is good but it may not be random enough.
We should mix them randomly if both are dominant or recessive. If the `firstAllele` always comes from the queen and the `secondAllele` always comes from the mate, it could allow the player to control the outcome.

src/main/java/forestry/core/genetics/Genome.java
```diff
@@ -56,17 +56,17 @@ private void checkChromosomes(IChromosome[] chromosomes) {
...
+						"Missing active allele for '%s'.\n%s", getSpeciesRoot().getUID(), chromosomeType.getName(), chromosomesToString(chromosomes));
 				throw new IllegalArgumentException(message);
 			}
 
-			IAllele secondary = chromosome.getSecondaryAllele();
+			IAllele secondary = chromosome.getInactiveAllele();
 			if (secondary == null) {
 				String message = String.format("Tried to create a genome for '%s' from an invalid chromosome template. " +
-						"Missing secondary allele for '%s'.\n%s", getSpeciesRoot().getUID(), chromosomeType.getName(), chromosomesToString(chromosomes));
+						"Missing inaktive allele for '%s'.\n%s", getSpeciesRoot().getUID(), chromosomeType.getName(), chromosomesToString(chromosomes));
```
mezz: `inaktive` should be `inactive`

src/main/java/forestry/core/utils/SimpleByteBuf.java
mezz: I still do not understand the bit logic here.

src/main/java/forestry/api/genetics/AlleleRegisterEvent.java
```diff
@@ -8,8 +8,9 @@
 import net.minecraftforge.fml.common.eventhandler.Event;
 
 /**
- * Called after Forestry has registered all his alleles of an specific type. Like IAlleleFruit or IAlleleSpecies
- * @since 5.3.3
+ * Called after Forestry has registered all of his alleles of an specific type. Like IAlleleFruit or IAlleleSpecies
```
mezz: `his` should be `its`

src/main/java/forestry/api/genetics/AlleleSpeciesRegisterEvent.java
```diff
@@ -6,7 +6,7 @@
 package forestry.api.genetics;
 
 /**
- * Called after Forestry has registered all his species alleles of a individual.
+ * Called after Forestry has registered all of his species alleles of a individual.
```
mezz: `his` should be `its`

src/main/java/forestry/api/genetics/ISpeciesRoot.java
```diff
@@ -104,7 +162,25 @@
...
+		return null;
+	}
+
 	/* TEMPLATES */
+	/**
+	 * @since Forestry 5.8
+	 */
+	IAlleleTemplateBuilder createTemplateBuilder();
```
mezz, line 176: Without `default` this will break Binnie's. 
If there is no way around it we can just update Binnie's, but then you probably don't need `default` on the `getWrapper` method above this.

mezz (changes requested):
Looking very good. I have some more comments for review on these new changes.

src/main/java/forestry/core/genetics/SaveFormat.java
```diff
@@ -0,0 +1,218 @@
...
+import net.minecraft.nbt.NBTTagList;
+
+import forestry.api.genetics.IAllele;
+import forestry.api.genetics.IChromosome;
+import forestry.api.genetics.IChromosomeType;
+import forestry.api.genetics.ISpeciesRoot;
+import forestry.core.utils.SimpleByteBuf;
+
+public enum SaveFormat {
+	//Used before forge fires the FMLLoadCompleteEvent.
```
mezz, line 15: Can you explain why this is needed? What is different before `FMLLoadCompleteEvent`?

src/main/java/forestry/core/genetics/Chromosome.java
Nedelosk: Do we not mix them randomly with `inheritChromosome()`?

src/main/java/forestry/core/genetics/SaveFormat.java
Nedelosk: The forge registry uses the ids that are saved in the `level.dat` file. All genetic items that we have created before the world was loaded would break, if we would use the `BINARY` format and the player would remove a mod that registers alleles. Because the items would use the ids from a version of the registry that has not the same ids like the one that was loaded from the world file.

src/main/java/forestry/core/utils/SimpleByteBuf.java
Nedelosk: I write a integer(`types`) to the byte array. I use it to check if the array contains the data of a specific chromsome type.
I create this int with `getValidChromosomes()`.

A small example:
I want to get the first chromosome (SPECIES) from a bee genome that I have writen to the byte array.
First I read `types`, it has the value 8191(1111111111111).
Now I wan't to know if this byte array contains this chromsome type. If it does not It is broken and has to be fixed.
I want the first chromsome so I have the index 0 and we have 13 bee chromosomes.
`1 << 13 - 0 - 1` = 4096 (1000000000000)
`1000000000000 & 1111111111111` = 1000000000000
`1000000000000 == 1000000000000` = true

src/main/java/forestry/api/genetics/ISpeciesRoot.java
Nedelosk: I have removed `default` from `getWrapper`. I don't think there is a way around it.

src/main/java/forestry/core/utils/SimpleByteBuf.java
mezz: How could we have a genome that is missing a chromosome type?

src/main/java/forestry/core/genetics/Chromosome.java
mezz: You're right, ok.

src/main/java/forestry/core/genetics/SaveFormat.java
mezz: That makes sense. I thought about the edge cases here and this is a nice solution.

src/main/java/forestry/core/utils/SimpleByteBuf.java
Nedelosk: Only if we add or remove a chromosome type I think. Should I remove this ?

mezz: Yeah I think this can be removed. 
We should avoid adding or removing chromosomes except on major breaking versions.

mezz:
I reverted one of my earlier changes and it caused a conflict with this PR. 
Can you update to resolve the conflicts?

mezz:
This looks very good to me but it's going to break addons.
Do you think we should wait for 1.13?

mezz (approved):

Issue #1923, opened 2017-12-22
BeeHives should burn with the forest


LemADEC:
As of 5.7.0.218, when a forest catches fires, the bee hives remains untouched, floating in mid air.
I would expect hives to be flammable and disappear, unless they are from a related 'nether' biome.

Issue #1710, opened 2017-06-15
1.10 Bees do damage on peaceful mode


Dilnu:
Hi, I'm not sure if this is intended or not but I take damage from bees on peaceful mode. This seems suboptimal.

alucard87pl:
Hey, maybe this could be an option in the configs?

PR #1930, opened 2017-12-30
Some minor additions

temp1011:
Resolves https://github.com/ForestryMC/ForestryMC/issues/1923

If you think some hives should not be flammable (maybe end?) or the burn values are wrong (I used values for logs) please comment below.

Hope this helps


temp1011:
Now resolves https://github.com/ForestryMC/ForestryMC/issues/1710 (and removes a todo I came across). I know these are weird issues to do but I'm not good enough at coding to be able to fix some of the more complicated bugs so I'm trying to help where I can. If the min and max values for ticksPerBeeWorkCycle are wrong please comment.

src/main/java/forestry/apiculture/worldgen/HiveDescription.java
```diff
@@ -146,6 +146,10 @@ public float getGenChance() {
 
 	@Override
 	public void postGen(World world, Random rand, BlockPos pos) {
+		if(isGoodBiome(world.getBiome(pos)))
+		{
+			Blocks.FIRE.setFireInfo(world.getBlockState(pos).getBlock(), 5, 5);
```
Nedelosk: This does not work like you think. `setFireInfo` sets the fire info for all forestry hives because all forestry hives have the same block instance. You have to implement the methods`getFlammability`, `getFireSpreadSpeed` and `isFlammable` in the class of the hive block.

Nedelosk (changes requested):

src/main/java/forestry/apiculture/worldgen/HiveDescription.java
temp1011: Ah OK. Makes sense.

temp1011:
Hopefully this should be the correct implementation now.

src/main/java/forestry/apiculture/blocks/BlockBeeHives.java
```diff
@@ -197,5 +199,30 @@ public EnumBlockRenderType getRenderType(IBlockState state) {
...
+	@Override
+	public boolean isFlammable(IBlockAccess world, BlockPos pos, EnumFacing face)
+	{
+		return !BiomeHelper.isBiomeHellish(world.getBiome(pos));
+	}
+	
+	@Override
+	public int getFireSpreadSpeed(IBlockAccess world, BlockPos pos, EnumFacing face)
+	{
+		if(!BiomeHelper.isBiomeHellish(world.getBiome(pos)))
```
Nedelosk: You don't have to test if the hive is in the nether, because the nether hive is from binnie and not from forestry.

Nedelosk (changes requested):

src/main/java/forestry/apiculture/blocks/BlockBeeHives.java
temp1011: ah, I didn't realise binnie had their own hive material and block

Nedelosk (approved):

PR #1938, opened 2018-01-06
Add power config for Analyzer

### LINKED ISSUE #1881: Issue: Analyzer block requires power
From the wiki https://forestryforminecraft.info/breeding:analyzer

The analyzer does not require power to operate but instead needs to be filled with liquid honey. 

However, the block erroneously (I assume) does actually require power.

temp1011:
Adds a config option to control power used per tick of the analyzer block to close https://github.com/ForestryMC/ForestryMC/issues/1881. 

I think this is ok for balance but if it isn't we can set a minimum or remove this. Also, if this is poorly implemented please remove.

Hope this helps.

### DIFF
--- a/src/main/java/forestry/core/tiles/TileAnalyzer.java
+++ b/src/main/java/forestry/core/tiles/TileAnalyzer.java
@@ -34,6 +34,7 @@ import forestry.api.arboriculture.TreeManager;
 import forestry.api.core.ForestryAPI;
 import forestry.api.genetics.AlleleManager;
 import forestry.api.genetics.IIndividual;
+import forestry.core.config.Config;
 import forestry.core.config.Constants;
 import forestry.core.errors.EnumErrorCode;
 import forestry.core.fluids.FilteredTank;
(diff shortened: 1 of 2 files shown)


src/main/java/forestry/core/tiles/TileAnalyzer.java
```diff
@@ -54,7 +55,7 @@
 public class TileAnalyzer extends TilePowered implements ISidedInventory, ILiquidTankTile, IItemStackDisplay {
 	private static final int TIME_TO_ANALYZE = 125;
 	private static final int HONEY_REQUIRED = 100;
-	private static final int ENERGY_PER_WORK_CYCLE = 20320;
+	private static final int ENERGY_PER_WORK_CYCLE = Config.analyzerEnergyPerWork;
```
Nedelosk: Please replace every appearance of `ENERGY_PER_WORK_CYCLE` with `Config.analyzerEnergyPerWork`. And remove `ENERGY_PER_WORK_CYCLE`.

Nedelosk (changes requested):

Nedelosk:
Thank you for your help

Nedelosk (approved):

temp1011:
Happy to help improve, unfortunately I'm not quite as good as you though :). I should really be thanking you for making forestry so great to play.

PR #1939, opened 2018-01-07
First go dimension config blacklisting

### LINKED ISSUE #1739: [1.12] [Suggestion] Dimension Blacklisting/Whitelisting for Worldgen
Hello,

I have a small suggestion to make for Forestry for Minecraft 1.12. Here's more detail on the matter:

### Issue Information 

- Minecraft 1.12
- Minecraft Forge 14.21.1.2046
- Forestry 5.4.3.129

### Issue Description

Within the common Forestry configuration file located in `.minecraft/config/forestry/common.cfg`  there are some basic values that you can use to alter the world generation, specific to enabling/disabling and retrogen. However, this is rather limiting especially with the lack of any other world generation manipulation mods such as CoFH World (prev. CoFH Core). 

The feature I'd like to see for world generation configuration for Forestry is dimension blacklisting. This would allow me (the user) to remove tree/hive/ore generation from other dimensions where you might not want Forestry trees/hives to spawn. E.g. Aroma1997's Dimension Mod, RFTools Dimensions, etc. 

I feel it would be a great addition to Forestry and it would be great to see in the mod in the future.

Thanks for creating and maintaining such a brilliant mod for all these years. 
Regards,
SnowShock35.

temp1011:
Aiming to close https://github.com/ForestryMC/ForestryMC/issues/1739. At the moment the config options at least show up in the .cfg file but I haven't tested fully because I need sleep. Also, this may be the wrong way to do things anyway. In any case config file and code may need some cleanup and localisations need adding. Feel free to edit if needed.

### DIFF
--- a/src/main/java/forestry/apiculture/worldgen/HiveDecorator.java
+++ b/src/main/java/forestry/apiculture/worldgen/HiveDecorator.java
@@ -45,8 +45,10 @@ public abstract class HiveDecorator {
 	}
 
 	public static void decorateHives(World world, Random rand, int chunkX, int chunkZ) {
-		if(!Config.canGenerateHives(world.provider.getDimension()))
+		if(!Config.canGenerateHives(world.provider.getDimension())) {
 			return;
+		}
+		
 		List<Hive> hives = ModuleApiculture.getHiveRegistry().getHives();
 
 		if (Config.generateBeehivesDebug) {
--- a/src/main/java/forestry/apiculture/worldgen/HiveDecorator.java
+++ b/src/main/java/forestry/apiculture/worldgen/HiveDecorator.java
@@ -45,6 +45,8 @@ public abstract class HiveDecorator {
 	}
 
 	public static void decorateHives(World world, Random rand, int chunkX, int chunkZ) {
+		if(!Config.canGenerateHives(world.provider.getDimension()))
+			return;
 		List<Hive> hives = ModuleApiculture.getHiveRegistry().getHives();
 
 		if (Config.generateBeehivesDebug) {
--- a/src/main/java/forestry/arboriculture/worldgen/TreeDecorator.java
+++ b/src/main/java/forestry/arboriculture/worldgen/TreeDecorator.java
@@ -40,6 +40,7 @@ import forestry.api.arboriculture.TreeManager;
 import forestry.api.genetics.AlleleManager;
 import forestry.api.genetics.IAllele;
 import forestry.arboriculture.commands.TreeGenHelper;
+import forestry.core.config.Config;
 import forestry.core.utils.BlockUtil;
 
 public class TreeDecorator {
@@ -54,6 +55,8 @@ public class TreeDecorator {
 	}
 
 	public static void decorateTrees(World world, Random rand, int worldX, int worldZ) {
+		if(!Config.canGenerateTrees(world.provider.getDimension()))
+			return;
 		if (biomeCache.isEmpty()) {
 			generateBiomeCache(world, rand);
 		}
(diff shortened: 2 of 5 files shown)


temp1011:
At the moment the config loads correctly but doesn't affect the worldgen from what I can see. I guess I'm missing something.

temp1011:
### Problems
- I think if I'm going to include a way to blacklist individual dimensions by ID I should include a way to blacklist dimensions by mod (e.g-no gen in Mystcraft ages or Rftools dimensions). However, at the moment it's quite awkward to implement because `world.getWorldInfo.getWorldName()` is the only way I can see to do this and I suspect quite a few mods will use useless names (`miningWorld` or something) for their dimensions.

- Furthermore, we can't guarantee dimension IDs between adding/removing mods or mods creating dimensions on the fly. (I'm betting some mods use the ID `DimensionManager.getNextFreeDimId()` for their ID).

- From what I can see, the current system makes it quite hard for other mods to keep track of which dimension is which.

### Proposal
- Add new `HashTable` (or maybe `BiHashMap`, could even use https://github.com/google/guava/wiki/NewCollectionTypesExplained#table) to the `DimensionManager` class in Forge to map between 'String' (the new input) and 'int' (dimension ID). 

- Registering dimensions by `int` will be deprecated and mods will be advised (maybe forced when using String?) to register dimensions by `"modName.dimensionName"`. `DimensionManager` will then handle assigning IDs. if a dimension if registered by  int ID it will have the `StringID` `"noModNameGiven.dimension" + int` or something. 

- It's then possible to have methods like `getDimensions(String modName)` which will allow us to blacklist mods from worldgen. 

- It also improves readability.

### Advantages
- Other mods can find dimensions even if not created at their usual dimension ID. 

- Forestry and any other mod with their own worldgen (pretty much most mods) can control where it generates much more easily.

- Dimensions abstracted away from meaningless numbers.

- Easier to register dimensions (slightly, don't have to worry about which IDs are free).

### Disadvantages
- System where some use `int` and some `String` to register dimensions could be messy. However, mods should be able to keep using deprecated methods to deal with 'int' or this could be introduced as a breaking change (1.13?)

- Mods will have to update their dimensions for this to have any use (making a breaking change in 1.13 would solve this)

- Probably a bunch more I can't think of.

### Possible further enhancments
- Allow `"modName.dimensionName.dimensionType"` as an extra argument so mods can make clear what type of dimension they have (eg - void world, mining world, etc)

**TLDR:** Some kind of dimension registry.

I guess this maybe should be in a PR/issue on the forge repo but I thought I would post here first to get see if this is a good idea or not.

@mezz @Nedelosk  What do you think? This is a pretty minor use for us but could be useful in general for most mods. Is this out of the scope of Forge or already implemented in a way I've missed? It definitely might be overthinking it for a config option PR :D.

Sorry for making you read so much but I thought this could be interesting.

mezz:
This seems good. I think it should also have a way for mods to disable the generation in their dimensions, by using IMC to tell Forestry.
I definitely prefer using names and mod names rather than numbers if possible, hopefully dimensions will be moving to use a real registry in 1.13 or so.

temp1011:
Should I close this in favour of https://github.com/ForestryMC/ForestryMC/pull/1941 which will solve things for now and then we can implement better worldgen if https://github.com/MinecraftForge/MinecraftForge/pull/4666 gets implemented?

Nedelosk:
@temp1011 Yes, but #1941 only contains a hive dimension blacklist. What about the tree and ore blacklist ?

Nedelosk:
@temp1011 I now have merged the other pull request.

temp1011:
We still need to do trees and ores. I can have a go at this. Hopefully it should be pretty similar to the other code.

IMarvinTPA:
@temp1011 anybody who fusses with Ores generally disables everything and uses something like COFH World to manage ore gen.  Just too many Copper/Tin/etc ores to deal with.
Trees on the other hand, that could use blacklisting since these trees are generally unique to Forestry.

temp1011:
Trees will definitely be blacklisted. It's nice to give users the option to do stuff inside forestry though

PR #1942, opened 2018-01-08
More Dimension Blacklists

temp1011:
This should hopefully sort out trees and ores. The code is pretty much the same as the other PR. Localisation PR will be done soon too.


src/main/java/forestry/arboriculture/worldgen/TreeDecorator.java
```diff
@@ -87,6 +88,8 @@ public static void decorateTrees(World world, Random rand, int worldX, int world
 	@Nullable
 	private static BlockPos getValidPos(World world, int x, int z, ITree tree){
 		// get to the ground
+		if(Config.blacklistedTreeDims.contains(world.provider.getDimension()))
```
Nedelosk: Please move this check to decorateTrees(World, Random, int, int). It makes not much sense to check it every time if the world is every time the same.

Nedelosk (changes requested):

Nedelosk (approved):

mezz (approved):

PR #1943, opened 2018-01-09
Add sorting block (Genetic Filter)

Nedelosk:
# The Genetic Filter
The **Genetic Filter** (please let me know if you have a better name) is mostly similar to the old **Apiarist's Pipe** but with some differences.
1. The block is no longer a buildcraft pipe that's why I had to change the way it transfers the items a little bit.  It only accepts items that it can output somewhere. Than it puts these items in a slot that is related to a side and pushes them to the tile on the side after a short delay (5 ticks). I had to add the delay so we don't have instant item transport.
2. Added a filter selection to the gui that has a search bar and a scroll bar and only appears if you select a filter slot. (The selected slot is marked with a icon)
3. Added 6 slot to the left side they are containing the buffed item of the side.
4. You can now use a bee / tree / butterfly to set the allele species filter. It uses the active allele if you press the left mouse button and the inactive if you press the right mouse button.
5. Added more filters related to trees and butterflies.
## Block Model
Currently not connected to pipes because there is no way to find out if a block is a pipe and if it is connected.

## The Item Model:
2018-01-09_02 24 29_genetic filter
## Recipes
```
[Wood, Diamond, Wood]
[Propolis, Glass, Propolis]
[Bronze Gear, Diamond, Bronze Gear]

[Wood, Diamond, Wood]
[Any Forestry Fruit, Glass, Any Forestry Fruit]
[Bronze Gear, Diamond, Bronze Gear]

[Wood, Diamond, Wood]
[Any Butterfly Caterpillar, Glass, Any Butterfly Caterpillar]
[Bronze Gear, Diamond, Bronze Gear]
```


Nedelosk:
Related to #396

AlexIIL:
I'm sorry to be a bit of a pain, but would it be possible to move the sorting rules (currently spread around ````TileGeneticFilter````) into a separate class that could be re-used by buildcraft compat to easily reimplement the functionality as a pipe? I'm not so sure about how the gui classes would work though (networking from forestry guis into buildcraft pipe tiles and vice-versa sounds like a massive headache), but I think separating out the core logic would be very useful for me.

Nedelosk:
I added a logic interface that you can use for buildcraft in the future.

AlexIIL:
@Nedelosk Thanks! That looks very useful now :)

AlexIIL:
I've implemented the propolis pipe in https://github.com/AlexIIL/BuildCraftCompat/tree/8.0.x-forestry-pipe, however in order to do this properly I'd like to use as much of the forestry implementation as possible (for example reusing the gui and container classes). I'm not sure what the best way of doing that is - in my private branch I've changed both ContainerGeneticFilter and GuiGeneticFilter to use an interface ````IFilterContainer````, which has a few methods to allow pipes to interact properly with it.

I can make a (proper, *without all of the hacks that my current branch has*) pull request with those changes if you want, but I'm not sure what would be best for forestry.

mezz (approved):

Nedelosk:
@AlexIIL https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/src/main/java/forestry/sorting/tiles/IFilterContainer.java

AlexIIL:
Thanks! I'll re-make my changes based off that :)

PR #1946, opened 2018-01-13
Removing overworked bees mechanic

### LINKED ISSUE #878: Add config option for the "overworked bees" productivity level
If pristine bees are overworked, they can become ignoble.
Right now it's hard-coded at 16x production, might as well make it configurable.

temp1011:
Should close https://github.com/ForestryMC/ForestryMC/issues/878. Based on the mezz's comment in that thread I've set max productivity (hopefully) to 16x and removed the behaviour which caused bees to degrade if they went past this production level.

If the decision has changed since then I can update/remove this. Hope this helps.

### DIFF
--- a/src/main/java/forestry/api/apiculture/IBeekeepingMode.java
+++ b/src/main/java/forestry/api/apiculture/IBeekeepingMode.java
@@ -37,6 +37,15 @@ public interface IBeekeepingMode {
 	 */
 	boolean isFatigued(IBee queen, IBeeHousing housing);
 
+	/**		
+	 * @return true if the queen is being overworked in the bee housing (with chance). will trigger a negative effect.		
+     * @deprecated bees max out at 16x productivity instead of being overworked.
+     * TODO remove this method in 1.13
+	 */
+	@Deprecated
+	 default boolean isOverworked(IBee queen, IBeeHousing housing) {
+		return false;
+		}
 	/**
 	 * @return true if the genetic structure of the queen is breaking down during spawning of the offspring (with chance). will trigger a negative effect.
 	 */
--- a/src/main/java/forestry/api/apiculture/IBeekeepingMode.java
+++ b/src/main/java/forestry/api/apiculture/IBeekeepingMode.java
@@ -38,11 +38,6 @@ public interface IBeekeepingMode {
 	boolean isFatigued(IBee queen, IBeeHousing housing);
 
 	/**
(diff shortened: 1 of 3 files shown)


src/main/java/forestry/api/apiculture/IBeekeepingMode.java
```diff
@@ -37,11 +37,6 @@
 	 */
 	boolean isFatigued(IBee queen, IBeeHousing housing);
 
-	/**
-	 * @return true if the queen is being overworked in the bee housing (with chance). will trigger a negative effect.
-	 */
-	boolean isOverworked(IBee queen, IBeeHousing housing);
```
mezz: To keep backward compatibility with mods that might use this method, you should deprecate this method like this:
```java
	/**		
	 * @return true if the queen is being overworked in the bee housing (with chance). will trigger a negative effect.		
     * @deprecated bees max out at 16x productivity instead of being overworked.
     * TODO remove this method in 1.13
	 */		
    @Deprecated
	default boolean isOverworked(IBee queen, IBeeHousing housing) {
        return false;
    }
```

mezz (changes requested):

src/main/java/forestry/api/apiculture/IBeekeepingMode.java
temp1011: Makes sense on retrospect not to break the interface 👍

src/main/java/forestry/api/apiculture/IBeekeepingMode.java
```diff
@@ -37,11 +37,15 @@
...
-	 * @return true if the queen is being overworked in the bee housing (with chance). will trigger a negative effect.
+	/**		
+	 * @return true if the queen is being overworked in the bee housing (with chance). will trigger a negative effect.		
+     * @deprecated bees max out at 16x productivity instead of being overworked.
+     * TODO remove this method in 1.13
 	 */
-	boolean isOverworked(IBee queen, IBeeHousing housing);
-
+	@Deprecated
+	 default boolean isOverworked(IBee queen, IBeeHousing housing) {
```
mezz: the indentation of this method is mixed tabs/spaces somehow

mezz (changes requested):

mezz:
Looks good, thanks!

mezz (approved):

PR #1948, opened 2018-01-14
Reintroduce the old-style farm and Rebalance electron tubes

Nedelosk:
# Reintroduce the old-style farm and Rebalance electron tubes
- [x] Reintroduce the old-style farms
- [x] Rebalance electron tubes


mezz (approved):
Looks very good, clean. Nice!

3TUSK:
Just a quick question, is there any chance that the old-style farm can place the soil as a ring (like what it used to do in 1.4.7 and before)?

Nedelosk:
@3TUSK I could add this as a config option, but I don't think that placing the soil as a ring is a good idea for the default pattern.

Nedelosk:
@3TUSK ?

3TUSK:
Yeah, the nostalgic style.

LuigiHutch:
Weren't harvesting/planting different blocks before? this just seems like a all in one magic block and a miniaturized multifarm

Nedelosk:
@LuigiHutch What would be the advantage of adding another block ?

LuigiHutch:
Previously, you needed one block for harvesting and one block for planting so you have to pipe between the two to create a completed setup.

Not really an advantage, makes it seem more interactive and you would have to put a bit more thought into your 'design'. This seems like 1 block does all, magic block too me.

This is just my opinion so feel free to disregard it :)

Nedelosk:
New Ender Farm:

Nedelosk:
@mezz Can you please take a second look, I changed a few things.

src/main/java/forestry/core/config/Config.java
```diff
@@ -104,6 +104,11 @@
 	private static boolean enableExUtilRedOrchid = true;
 	private static boolean enableMagicalCropsSupport = true;
 
+	// Cultivation
+	public static int planterExtend = 4;
+	public static boolean ringFarms = false;
```
mezz: I think the farms should be rings by default, it makes them visually different from other mods and it has nostalgic appeal for a lot of players.

src/main/java/forestry/cultivation/tiles/TilePlanter.java
```diff
@@ -0,0 +1,510 @@
...
+		BlockPos coords = getCoordinates();
+		return ClimateUtil.getHumidity(getWorldObj(), coords);
+	}
+
+	@Override
+	public ITankManager getTankManager() {
+		return tankManager;
+	}
+
+	public abstract ItemStack[] createGermlingStacks();
```
mezz: I think these three should be `NonNullList<ItemStack>`

src/main/java/forestry/farming/FarmRegistry.java
```diff
@@ -60,64 +65,132 @@ public void registerFarmables(String identifier, IFarmable... farmablesArray) {
...
+
+		return fertilizerMap.build();
+	}
+
+	private Map<String, String> parseConfig(String[] fertilizerList, ImmutableMap.Builder<ItemStack, Integer> fertilizerMap){
+		Map<String, String> configEntries = new HashMap<>();
+		for(String entry : fertilizerList){
+			String[] spited = entry.split(";");
+			if(spited.length < 2){
+				Log.error("Foretsry failed to parse a entry of the fertilizer config.");
```
mezz: >Foretsry

typo

src/main/java/forestry/farming/FarmRegistry.java
```diff
@@ -60,64 +65,132 @@ public void registerFarmables(String identifier, IFarmable... farmablesArray) {
...
+				continue;
+			}
+			String itemName = spited[0];
+			ItemStack fertilizerItem = ItemStackUtil.parseItemStackString(itemName, OreDictionary.WILDCARD_VALUE);
+			if(fertilizerItem == null || fertilizerItem.isEmpty()){
+				Log.error("Foretsry failed to parse a entry of the fertilizer config, because the item doesn't exists.");
```
mezz: >Foretsry

typo

src/main/java/forestry/farming/logic/FarmLogicCrops.java
```diff
@@ -96,20 +85,17 @@ private boolean trySetCrop(World world, IFarmHousing farmHousing, BlockPos posit
...
-				}
-			}
+	public ItemStack getIconItemStack() {
+		return new ItemStack(Items.WHEAT);
+	}
+
+	@Override
+	public String getName() {
+		if (isManual) {
+			return "Manual Crop Farm";
```
mezz: Do these need localization?

src/main/java/forestry/farming/logic/FarmProperties.java
```diff
@@ -0,0 +1,87 @@
...
+
+	@Override
+	public void registerFarmables(String identifier) {
+		farmablesIdentifiers.add(identifier);
+	}
+
+	@Override
+	public Collection<IFarmable> getFarmables() {
+		if (farmables == null) {
+			farmables = farmablesIdentifiers.stream().map(FarmRegistry.getInstance()::getFarmables).flatMap(Collection::stream).collect(Collectors.toSet());
```
mezz: this is really long so it should probably have newlines:
```java
farmables = farmablesIdentifiers.stream()
    .map(FarmRegistry.getInstance()::getFarmables)
    .flatMap(Collection::stream)
    .collect(Collectors.toSet());
```

mezz (changes requested):
This looks very good, I only found some small things.

src/main/java/forestry/farming/logic/FarmLogicCrops.java
3TUSK: If they are actually showing in any GUI, please do so.

src/main/java/forestry/cultivation/gui/widgets/GhostItemStackWidget.java
```diff
@@ -30,25 +30,27 @@ public void draw(int startX, int startY) {
 		if (!slot.getHasStack()) {
 			super.draw(startX, startY);
 		}
-		GL11.glDisable(GL11.GL_LIGHTING);
-		GL11.glDisable(GL11.GL_DEPTH_TEST);
-		GL11.glEnable(GL11.GL_BLEND);
+		GlStateManager.pushMatrix();
```
mezz: `pushMatrix` and `popMatrix` are only needed when there is a GL transform, but I don't see any in the code below.

mezz (changes requested):
Looking good, I can only find one thing to comment on in these new changes.

KorDum:
"New" multifarm will gone?

Nedelosk:
No, Why should I remove the multifarm ? It's a very good and flexible farm.

KorDum:
Sorry, and why then return the old, if it is, in fact, the same thing? Only much cheaper in crafting.

Nedelosk:
It's much easier to use for new users. And it is not the same like the multifarm.It's smaller, cheaper and only contains one farm logic.

Nedelosk:
And we can test new things with the new farm without breaking the code of the multifarm.

Nedelosk:
@KorDum You don't have to be sorry. You have done  nothing  wrong. You only asked a question.

KorDum:
Thanks for answer!
But it seems to me, then in multifarm there is no sense. I can put a few cheap "old" farms on everything I need, and that's it, right?

KorDum:
From my own experience, I know that players will always look for something that is easier. If multifarm is more expensive, but performs exactly the same, then the priorities of the players will shift.

Nedelosk:
If you add more gearboxes to the mulifarm, it is much faster than a one of the new farms and your farmland is much bigger.

KorDum:
Is it possible to make the old farm consume more fertilizers?

Nedelosk:
I should add a fertilizer modifier to the config and double it for the new farms.

Nedelosk:
And I probably add a new casing for the new farms to make them more expensive.

KorDum:
Looks like cool!

KorDum:
Maybe configurable (in config) side size (or radius from center) for old farm? 15 by default.

Nedelosk:
@mezz I removed `pushMatrix` and `popMatrix`, everything should be correct now.

src/main/java/forestry/farming/multiblock/FarmController.java
```diff
@@ -558,7 +558,7 @@ private boolean cullCrop(ICrop crop, IFarmLogic provider) {
 			}
 		}
 
-		final int fertilizerConsumption = provider.getFertilizerConsumption();
+		final int fertilizerConsumption = (int)(provider.getFertilizerConsumption() * Config.fertilizerModifier);
```
mezz: I think in some cases it may be important to use `Math.round(` instead of casting to `(int)` for these. Better to use it and be more correct.

mezz (changes requested):
Looks good, I didn't find any big issues.

src/main/java/forestry/farming/multiblock/FarmController.java
Nedelosk: https://github.com/ForestryMC/ForestryMC/pull/1948/commits/05d1eb87ce1132fb82e2c60d10324697ab2d44da

mezz (approved):

PR #1965, opened 2018-02-02
Add IMC messages for dimension blacklist

temp1011:
As discussed in https://github.com/ForestryMC/ForestryMC/pull/1939 mods should have a way to tell Forestry to not generate bees/trees/ore in their dimension. Hopefully this PR will add that functionality using IMC. I haven't used IMC before so let me know if I have gone wrong.


mezz (approved):
Looks good to me

PR #1967, opened 2018-02-03
Add config for ignoble decay

### LINKED ISSUE #1409: onPostQueenDeath method
Hello. I'd like to ask you to add onPostQueenDeath method to BeeListener. This feature was in forestry 3

temp1011:
Should close https://github.com/ForestryMC/ForestryMC/issues/1409. I don't know if this breaks balance too much or not. However, this is simpler than adding in `onPostQueenDeath()`.

### DIFF
--- a/src/main/java/forestry/apiculture/ModuleApiculture.java
+++ b/src/main/java/forestry/apiculture/ModuleApiculture.java
@@ -904,6 +904,10 @@ public class ModuleApiculture extends BlankForestryModule {
 	public static double getSecondPrincessChance() {
 		return secondPrincessChance;
 	}
+	
+	public static void setAlvearyDecay(boolean decay) {
+		ignobleDecayInAlveary = decay;
+	}
 
 	private static void parseBeeBlacklist(String[] items) {
 		for (String item : items) {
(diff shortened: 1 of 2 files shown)


mezz:
I would prefer to have this config be more granular, with something like `ignoble generations = 100`, and `ignoble decay chance = 0.02`.
If we add the config you have here, I think it would make more sense to make every bee pristine so it is less confusing to players.

SandGrainOne:
This PR doesn't answer #1409 either in my opinion. They wanted a way to prevent the death of a single bee under the right conditions. Meaning the player has worked for it and added something to the beeHouse (I presume).

temp1011:
I agree this isn't specific enough at the moment. I'll change this at some point to reflect the comments better.

temp1011:
Ok I've changed how this works a bit. There's `boolean` in `ModuleApiculture` that controls whether an ignoble bee in an alveary will decay. There's also a general option for setting the genetic decay chance of an ignoble bee. Hopefully this addresses things better but I'm still interested to see if I should be doing things differently.
@NightKosh what do you think?

temp1011:
This probably doesn't need to be open...

PR #1968, opened 2018-02-03
Add whitelist for generation based on dimension

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/1739. I don't think it's possible to change where the Apiarist village house will generate easily so I haven't touched that. However, I think it is possible to implement a whitelist as well for dimensions. This is designed for pack makers/players (eg - player has access to many dimensions early but pack maker wants to gate forestry until later) as the blacklist will take priority (as it should do in my opinion).

However, this is slightly messy. If there's a better way to do something like this let me know.

### DIFF
--- a/src/main/java/forestry/apiculture/HiveConfig.java
+++ b/src/main/java/forestry/apiculture/HiveConfig.java
@@ -23,11 +23,16 @@ public class HiveConfig {
 	private final Set<Biome> blacklistedBiomes = new HashSet<>();
 
 	private static final Set<Integer> blacklistedDims = new HashSet<>();
+	
+	private static final Set<Integer> whitelistedDims = new HashSet<>();
 
 	public static void parse(LocalizedConfiguration config){
 		config.addCategoryCommentLocalized(CATEGORY);
 		for(int dimId : config.get("world.generate.beehives", "dimBlacklist", new int[0]).getIntList()){
 			blacklistedDims.add(dimId);
+		}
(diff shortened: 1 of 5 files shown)


src/main/java/forestry/apiculture/HiveConfig.java
```diff
@@ -62,11 +67,18 @@ private boolean isBlacklisted(Biome biome){
 		return BiomeDictionary.getTypes(biome).stream().anyMatch(blacklistedTypes::contains);
 	}
 	
-	public static boolean isDimBlacklisted(int dimId) {
-		return blacklistedDims.contains(dimId);
+	public static boolean isDimAllowed(int dimId) {		//blacklist has priority
+		if(blacklistedDims.isEmpty() || !blacklistedDims.contains(dimId)) {
+			return whitelistedDims.isEmpty() || whitelistedDims.contains(dimId);
+		}
+		return true;
```
Nedelosk: This should be `false`.

Nedelosk: This should be `false`.

Nedelosk (changes requested):

src/main/java/forestry/apiculture/HiveConfig.java
temp1011: done

Nedelosk (approved):

mezz (approved):

PR #1987, opened 2018-02-11
Add IMC to let mods add fertilizer

temp1011:
This is more of an idea but since the PR is so easy to write I went ahead and implemented it.

 This PR adds a way for mods to add their fertilizers to work in a farm. There is already a config to do this so I don't think it's too unreasonable. However, this relies on some good faith in other mods not to add something which will break the balance of Forestry. This also means mods don't have to write a plugin and can just send a message instead.

If things just stay in the config then at least the only people who will add fertilizers are players dedicated enough to actually change default configs (!) or packmakers. This is probably good because these people will care about balance more.

Let me know what you think.


temp1011:
This should be done from the other mod's side so they actually are willing to put some thought into this I think. Also this code is wrong I'm pretty sure.

Issue #1210, opened 2016-05-19
Look into adding Apiarist Villager AI


mezz:
The bees are not automated, the apiarist villagers should remove bees and replace them.

PR #2001, opened 2018-02-22
Add AI for Apiarist to Replace Bees in Village Bee Houses

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/1210.

This PR adds AI to the apiarist villager to replace bees in the bee houses found in villages. The villager will replace bees from the beehouse first and then use those in their inventory. If there are any bees left in the beehouse at the end the villager will then add them to it's internal inventory.

At the moment the villager does nothing with combs/other products and will do nothing if the inventory in the beehouse or their inventory becomes full. However, this should only occur if the villager is working in loaded chunks which means a player should be able to do some maintenance. I'm happy to add stuff for this in though.

The code is quite verbose so any suggestions on making it cleaner are welcome.

### DIFF
--- a/src/main/java/forestry/apiculture/ApiaristAI.java
+++ b/src/main/java/forestry/apiculture/ApiaristAI.java
@@ -58,7 +58,7 @@ public class ApiaristAI extends EntityAIMoveToBlock {
 		BlockPos housePos = this.destinationBlock.north().up();
 		this.villager.getLookHelper().setLookPosition(housePos.getX() + 0.5D, housePos.getY(), housePos.getZ() + 0.5D, 10.0F, this.villager.getVerticalFaceSpeed());
 
-		if (this.villager.getDistanceSqToCenter(housePos) < 2.0D) {
+		if (this.getIsAboveDestination()) {
 			World world = this.villager.world;
 
 			TileBeeHouse beeHouse = (TileBeeHouse) TileUtil.getTile(world, housePos);
--- a/src/main/java/forestry/apiculture/ApiaristAI.java
+++ b/src/main/java/forestry/apiculture/ApiaristAI.java
@@ -147,7 +147,6 @@ public class ApiaristAI extends EntityAIMoveToBlock{
 					if(type==EnumBeeType.PRINCESS) foundPrincess = true;
 					if(type==EnumBeeType.DRONE) foundDrone = true;
 					if(foundDrone && foundPrincess) {
-						System.out.println("Found bees!");
 						return true;
 					}
 				}
--- a/src/main/java/forestry/apiculture/ApiaristAI.java
+++ b/src/main/java/forestry/apiculture/ApiaristAI.java
@@ -1,0 +1,153 @@
+package forestry.apiculture;
+
+import forestry.api.apiculture.EnumBeeType;
+import forestry.apiculture.blocks.BlockApiculture;
+import forestry.apiculture.items.ItemBeeGE;
+import forestry.apiculture.tiles.TileBeeHouse;
+import forestry.core.tiles.TileUtil;
+import forestry.core.utils.InventoryUtil;
+import net.minecraft.block.Block;
+import net.minecraft.entity.ai.EntityAIMoveToBlock;
+import net.minecraft.entity.passive.EntityVillager;
+import net.minecraft.inventory.InventoryBasic;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@SuppressWarnings("static-access")
+public class ApiaristAI extends EntityAIMoveToBlock{
+	private final EntityVillager villager;
+	private boolean hasDrone;
+	private boolean hasPrincess;
+	private InventoryBasic inventory;
+	
+	public ApiaristAI(EntityVillager villager, double speed) {
+		super(villager, speed, 16);
+		this.villager = villager;
+		inventory = villager.getVillagerInventory();
+	}
+	
+	@Override
+	public boolean shouldExecute() {
+		if(this.runDelay<0) {
+			this.hasDrone=hasBeeType(EnumBeeType.DRONE);
+			this.hasPrincess=hasBeeType(EnumBeeType.PRINCESS);
+		}
+		return super.shouldExecute();
+	}
+	
+	@Override 
+	public void updateTask()
+	 {
+        super.updateTask();
+        this.villager.getLookHelper().setLookPosition((double)this.destinationBlock.getX() + 0.5D, (double)(this.destinationBlock.getY() + 1), (double)this.destinationBlock.getZ() + 0.5D, 10.0F, (float)this.villager.getVerticalFaceSpeed());
+
+        if (this.getIsAboveDestination())
+        {
+            World world = this.villager.world;
+            BlockPos blockpos = this.destinationBlock;
+            if(!(world.getBlockState(blockpos).getBlock() instanceof BlockApiculture) || !(TileUtil.getTile(world, blockpos) instanceof TileBeeHouse)) {
+            	return;
+            }
+        	TileBeeHouse beeHouse = (TileBeeHouse) TileUtil.getTile(world,blockpos);
+			InventoryBeeHousing inventory = (InventoryBeeHousing) beeHouse.getBeeInventory();
+            //fill princess slot from beehouse inventory
+            if(inventory.getStackInSlot(inventory.SLOT_QUEEN).isEmpty()) {
+            	for(int i = inventory.SLOT_PRODUCT_1 ; i < inventory.SLOT_PRODUCT_COUNT ; i++) {
+    				if(inventory.getStackInSlot(i)!=ItemStack.EMPTY && inventory.getStackInSlot(i).getItem() instanceof ItemBeeGE) {
+    					if(((ItemBeeGE)inventory.getStackInSlot(i).getItem()).getType()==EnumBeeType.PRINCESS) {
+    						inventory.setQueen(inventory.getStackInSlot(i).copy());
+    						inventory.setInventorySlotContents(i, ItemStack.EMPTY);
+    						break;
+    					}
+    				}
+    			}
+            }
+            //fill drones from beehouse inventory (these stack so more work needed)
+            for(int i = inventory.SLOT_PRODUCT_1 ; i < inventory.SLOT_PRODUCT_COUNT ; i++) {
+            	if(inventory.getStackInSlot(i)!=ItemStack.EMPTY && inventory.getStackInSlot(i).getItem() instanceof ItemBeeGE) {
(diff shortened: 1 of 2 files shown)


src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
+package forestry.apiculture;
```
mezz, line 12: Please add the copyright header at the top

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+import forestry.core.utils.InventoryUtil;
+import net.minecraft.block.Block;
+import net.minecraft.entity.ai.EntityAIMoveToBlock;
+import net.minecraft.entity.passive.EntityVillager;
+import net.minecraft.inventory.InventoryBasic;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+@SuppressWarnings("static-access")
```
mezz: This is not needed, you should fix the static access warnings instead of suppressing them.

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+public class ApiaristAI extends EntityAIMoveToBlock{
```
mezz: Please fix the formatting for spacing around braces and parenthesis in this file.
There are formatting settings for the project here: https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/formatting-settings.jar

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+			this.hasPrincess=hasBeeType(EnumBeeType.PRINCESS);
+		}
+		return super.shouldExecute();
+	}
+	
+	@Override 
+	public void updateTask()
+	 {
+        super.updateTask();
+        this.villager.getLookHelper().setLookPosition((double)this.destinationBlock.getX() + 0.5D, (double)(this.destinationBlock.getY() + 1), (double)this.destinationBlock.getZ() + 0.5D, 10.0F, (float)this.villager.getVerticalFaceSpeed());
```
mezz: casting to double and float here is not needed

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+
+        if (this.villager.getDistanceSqToCenter(this.destinationBlock) > 4.0D)
+        {
```
mezz: one file should not have multiple brace styles. please use the brace style of this project

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+            World world = this.villager.world;
+            BlockPos blockpos = this.destinationBlock;
+            if(!(TileUtil.getTile(world, blockpos) instanceof TileBeeHouse)) {
+            	return;
+            }
+        	TileBeeHouse beeHouse = (TileBeeHouse) TileUtil.getTile(world,blockpos);
```
mezz: There are mixed tabs and spaces for indentation in many places

mezz: You can avoid calling `getTile`  two times by doing this:
```java
TileBeeHouse beeHouse = TileUtil.getTile(world, blockpos, TileBeeHouse.class);
if (beeHouse == null) {
	return;
}
```

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+            
+			//fill slots from inside bee house
+        	for(ItemStack stack : InventoryUtil.getStacks(inventory, inventory.SLOT_PRODUCT_1, inventory.SLOT_PRODUCT_COUNT)) {
+        		if(stack!=ItemStack.EMPTY && stack.getItem() instanceof ItemBeeGE) {
+        			EnumBeeType type = ((ItemBeeGE)stack.getItem()).getType();
+        			if(inventory.getStackInSlot(inventory.SLOT_QUEEN).isEmpty() && type==EnumBeeType.PRINCESS) {
+        				inventory.setQueen(stack.copy());
+						stack.setCount(0);
+        			} else if(type==EnumBeeType.DRONE) {
+        				InventoryUtil.addStack(inventory, stack, inventory.SLOT_DRONE, 1, true);
```
mezz: I think this logic will duplicate the drone. You should check the return value of this method and shrink the stack by that amount.

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,152 @@
...
+        	
+        	//fill slots from villager inventory
+        	if(inventory.getStackInSlot(inventory.SLOT_DRONE).isEmpty() || inventory.getStackInSlot(inventory.SLOT_QUEEN).isEmpty()) {
+        		boolean princessAdded = false;
+        		boolean droneAdded = false;
+        		for(ItemStack stack : InventoryUtil.getStacks(villagerInventory)) {
+        			if(princessAdded && droneAdded) {
+        				break;
+        			}
+        			if(stack!=ItemStack.EMPTY && stack.getItem() instanceof ItemBeeGE) {
```
mezz: Never compare `ItemStack.EMPTY` with `==` or `!=`. You must use `!stack.isEmpty()` here.

src/main/java/forestry/core/EventHandlerCore.java
```diff
@@ -74,7 +79,7 @@ public void handlePlayerChangedDimension(PlayerEvent.PlayerChangedDimensionEvent
 		syncBreedingTrackers(player);
 	}
 
-	private static void syncBreedingTrackers(EntityPlayer player) {
+	private void syncBreedingTrackers(EntityPlayer player) {
```
mezz: This method should not need to be changed.

src/main/java/forestry/core/EventHandlerCore.java
```diff
@@ -140,4 +145,16 @@ public void lootLoad(LootTableLoadEvent event) {
...
 		}
 	}
+	
+	@SubscribeEvent
+	public void handleVillagerAI(EntityJoinWorldEvent event) {	//won't register if villager not registered
+		IForgeRegistry<VillagerRegistry.VillagerProfession> villagerProfessions = ForgeRegistries.VILLAGER_PROFESSIONS;
+		Entity entity = event.getEntity();
+		if((entity instanceof EntityVillager)) {
+			EntityVillager villager = (EntityVillager) event.getEntity();
+			if(villagerProfessions.getKey(villager.getProfessionForge()).toString().equals(Constants.ID_VILLAGER_APIARIST)) {
```
mezz: Please break this long line into smaller parts by storing things to variables before the `if`.

mezz (changes requested):
Looks good but needs some cleanup and a close look at some of the item handling to prevent duplication.

mezz:
During testing, the villager seemed to replace the bee while they were inside the house so they were nowhere to be seen when it happened.

src/main/java/forestry/apiculture/ApiaristAI.java
temp1011: Yep I missed this one

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,170 @@
...
+
+			TileBeeHouse beeHouse = (TileBeeHouse) TileUtil.getTile(world, blockpos);
+			if (beeHouse == null) {
+				return;
+			}
+			InventoryBeeHousing inventory = (InventoryBeeHousing) beeHouse.getBeeInventory();
+
+			//fill slots from inside bee house
+			for (ItemStack stack : InventoryUtil.getStacks(inventory, SLOT_PRODUCT_1, SLOT_PRODUCT_COUNT)) {
+				if (stack != ItemStack.EMPTY && stack.getItem() instanceof ItemBeeGE) {
```
mezz: You must use `!stack.isEmpty()` here.

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,170 @@
...
+					return false;
+				}
+			}
+			boolean foundPrincess = hasPrincess;
+			boolean foundDrone = hasDrone;
+			if (foundDrone && foundPrincess) {
+				return true;
+			}
+			for (ItemStack stack : InventoryUtil.getStacks(inventory, SLOT_PRODUCT_1, SLOT_PRODUCT_COUNT)) {
+				if (stack != ItemStack.EMPTY && stack.getItem() instanceof ItemBeeGE) {
```
mezz: You must use `!stack.isEmpty()` here.

mezz (changes requested):

temp1011:
Movement to bee house broken at the moment since the villager doesn't like climbing on top of it.

temp1011:
Hopefully this should fix things. The block to the north of bee houses should be fine for the villager to move to so they should be able to get to the bee house now.

temp1011:
Still no luck with getting the villager to move anywhere. Space should be free for them to move to and they'll look at bee house/replace bees if close enough...

mezz:
Villagers move to crops I think, or maybe they just randomly wander over to them and the AI becomes active? I'm not sure. In any case that logic may be a good place to look.

temp1011:
I've tried to base this logic on that one as much as possible. I'll have another look now. It could just be that the villager takes a long time to scan for the block to move to.

temp1011:
Fixed movement. Villagers always move to north side of bee house but this shouldn't be a problem in villages.

src/main/java/forestry/apiculture/ApiaristAI.java
```diff
@@ -0,0 +1,171 @@
...
+
+	public ApiaristAI(EntityVillager villager, double speed) {
+		super(villager, speed, 16);
+		this.villager = villager;
+		villagerInventory = villager.getVillagerInventory();
+	}
+
+	@Override
+	public boolean shouldExecute() {
+		if (this.runDelay < 0) {
```
mezz: `this.runDelay` is never less than 0 so this is never true.
Please test this more carefully.

mezz (changes requested):

temp1011:
Ahhhh sorry. I copied the code out of `EntityAIHarvestFarmland` wrong...

mezz:
Looks good now. I think the range may need to be tweaked later, the villager isn't really close enough to the bees most of the time so this logic isn't active.

mezz (approved):

Issue #1625, opened 2017-03-13
un-pollinated leaves should breed with themselves


EnderiumSmith:
when a leaf block that has not been pollinated drops a sapling the sapling genetics should be generated by breeding that leaf block with itself. this way its easier to get saplings with homozigous traits and allows to shuffle the traits without waiting for the bees to pollinate them. do bees even pollinate identical leaf blocks? as a trade off mutations should not be generated this way so you still have to use bees for new species.

mezz:
That's an interesting idea. We'll have to think about the implications a little bit more before implementing it though.
Bees do not pollinate identical leaves together, mainly to speed up the pollination process by avoiding "wasting" a pollination attempt that has no chance of mutation.

PR #2006, opened 2018-02-24
Add self-pollination of leaves near bees

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/1625.

I've added self pollination of leaves around bee hives in the same way as normal pollination occurs (but without the hive collecting pollen). It is configurable but is on by default. I took the chance of it down but I'm happy to tweak numbers. The chance is separate so this won't slow down normal pollination.

### DIFF
--- a/src/main/java/forestry/api/apiculture/IBee.java
+++ b/src/main/java/forestry/api/apiculture/IBee.java
@@ -114,10 +114,5 @@ public interface IBee extends IIndividualLiving {
 	IIndividual retrievePollen(IBeeHousing housing);
 
 	boolean pollinateRandom(IBeeHousing housing, IIndividual pollen);
-	
-	//TODO - implement fully in 1.13
-	default void selfPollinate(IBeeHousing housing) {
-		return;
-	}
 
 }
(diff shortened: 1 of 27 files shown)


Nedelosk:
That was not the idea behind #1625. The idea was to breed the sapling with itself at the moment the player breaks the leaf.

temp1011:
I thought this would be better otherwise a player could just bonemeal and break a tree to breed it without any need for bees. However if this is wrong please close

mezz:
I do not think I understand this implementation.
Why is this copied from the existing pollination code, and not just part of it?
The idea of bees pollinating the leaves makes sense to me though.

mezz (changes requested):
I don't think this logic is quite right.
The check for self-pollination is in `TileLeaves#canMateWith` and in `CheckPollinatableTree#canMateWith`.
I think you can alter the logic in those two places so that it looks at `ModuleApiculture.doSelfPollination` and skips the check for `isGeneticEqual`.

temp1011:
Yep that's much nicer, sorry.

mezz:
Cool, looks good to me.

mezz (approved):

PR #2013, opened 2018-02-26
Run Formatting on src/main/java

temp1011:
Created by running format code on src/main/java. Not hugely necessary but nice to have consistent code style and clean things up. Hopefully these aren't just problems introduced by my IDE.

Formatting settings taken from https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/formatting-settings.jar.


temp1011:
Realistically this needs someone with direct commit privileges to do.

PR #2024, opened 2018-03-06
Improve Compatibility Support

### LINKED ISSUE #2019: [Suggestion] Compatability with TechReborn Rubber Trees in Farm?
Would it be possible if the Rubber Plantation mode in the Forestry Farm multiblock also supported the TechReborn trees (only seems to support IC2 Trees from what I can see ingame)?

TechReborn - 2.12.3.477
Forestry - 5.8.0.250

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2019.

- move some null checks to make code cleaner.

- begin work on farmable bushes from natura, biomesoplenty etc.

- fix biomes o plenty plugin.

- improve industrial craft plugin.

- remove chisel plugin, it's not needed from what I an see, comment if still needed.

- checked ender io plugin against beta. Still works, may change since still in beta.


Nedelosk (changes requested):
Could you please use the new Module Helper that I added with the last commit.

temp1011:
I think this needs to be split up completely anyway.

Issue #2022, opened 2018-03-04
[Suggestion] Add Agricraft cactus to accepted flower type


SjetCraft:
With Agricraft disabling vanilla farming (via config) cactus becomes unplaceable. 

I suggest adding compatibility with Agricraft cactus cropsticks as a valid flower type for Modest bees.

temp1011:
https://github.com/ForestryMC/ForestryMC/issues/1698 would also be a solution.

PR #2027, opened 2018-03-08
 Add IMC for registering flowers

temp1011:
This PR lets mods register their own flowers with forestry. This works a lot better than doing it through plugins on the forestry side since another mod has much better access to it's own blockstates.

For example, in https://github.com/ForestryMC/ForestryMC/issues/2022 rather than forestry checking when an agricraft config option is disabled then finding the correct blockstate and registering it there could be a plugin for Forestry in agricraft to just send an imc if the config option is changed (I would be happy to write this plugin).

### DIFF
--- a/src/main/java/forestry/apiculture/ModuleApiculture.java
+++ b/src/main/java/forestry/apiculture/ModuleApiculture.java
@@ -995,7 +995,7 @@ public class ModuleApiculture extends BlankForestryModule {
 				return false;
 			}
 		}
-		
+
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/forestry/apiculture/ModuleApiculture.java
```diff
@@ -988,6 +990,39 @@ public boolean processIMCMessage(IMCMessage message) {
 				HiveConfig.addBlacklistedDim(dim);
 			}
 			return true;
+		} else if (message.key.equals("add-plantable-flower")) {
+			try {
```
mezz: This method is getting very long, can you separate out these new cases into new methods?

mezz (changes requested):
Looks good, I like the idea.

temp1011:
Should the `try/catch` be inside or outside the method call? Inside looks a bit weird but outside increases the other method length.

mezz:
Inside is fine I think.

mezz (approved):

Issue #2030, opened 2018-03-10
[Minor] Block of Charcoal does not register with Ore Dictionary


ghost:
The item `forestry:charcoal` should register with the Forge Ore Dictionary as `blockCharcoal` so it can be treated equally with the equivalent Actually Additions, Chisel, and Garden Stuff blocks.

temp1011:
Looks like it's there but not used.

PR #2032, opened 2018-03-11
Register Charcoal Block with Ore Dictionary

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2030. Hopefully this is fine.


temp1011:
~~Needs this sorting first:
`[19:14:28] [main/WARN] [FML]: * A broken ore dictionary registration with name blockCharcoal has occurred. It adds an item (type: class forestry.arboriculture.blocks.BlockRegistryCharcoal$1) which is currently unknown to the game registry. This dictionary item can only support a single value when registered with ores like this, and NO I am not going to turn this spam off. Just register your ore dictionary entries after the GameRegistry.
TO USERS: YES this is a BUG in the mod Forestry report it to them!`~~

EDIT: Fixed

Issue #2008, opened 2018-02-25
Make the blocks/items can be crated configurable


LuigiHutch:
- Forge: 14.23.2.2618
- Forestry: 5.8.0.250

Another feature request, add a config which allows you to choose which blocks/items can be crated.

Im not sure how feasible this would be but hopfully you will consider it.

Kind regards,
LuigiHutch

bookerthegeek:
Would this allow us to set any item to be able to be put into a crate?

PR #2047, opened 2018-03-18
Add configuration for crates

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2008. Needs localization but that should essentially be the same as the backpack config apart from the fact we don't display the default items in crates here.


temp1011:
https://github.com/temp1011/ForestryMC/blob/2adaacdc04f57958d03366da39c834068de53542/src/main/java/forestry/storage/ModuleBackpacks.java#L391-L404

I'm assuming I should add this to forge? Recipes are blacklisted but other than that would be nice to have.

temp1011:
https://github.com/MinecraftForge/MinecraftForge/pull/4807

temp1011:
PR to Forge was merged so we don't need to register fences and fence gates anymore.

temp1011:
Fixed conflicts

mezz:
To make reviewing easier, please try to keep commits to one small subject.
This one can be broken into:

1. Add configuration for crates
2. Formatting fixes (on unrelated files)
3. Remove oredict registration for fences and fence gates

If there is a problem with any one of these, the PR gets stalled and it makes changes harder to review. If they are separate, I can merge the good ones and comment on the one that needs fixing.

Review:

We shouldn't remove the oredict registration without increasing the minimum forge version.
Now that localization is merged into the main project here, these configs need localization in the same PR.

temp1011:
I've bumped to the latest recommended forge version which includes the commit to add fences to the forge oredict. It would be good if someone with direct commit privileges could format the whole code base. I could open a PR but it would get outdated quite quickly.

gradle.properties
```diff
@@ -1,5 +1,5 @@
 mcversion=1.12.2
-forgeversion=14.23.2.2637
+forgeversion=14.23.3.2655
```
mezz: This is the forge version used in development. 
It's good to update this, but we also need to update the minimum required version here: https://github.com/ForestryMC/ForestryMC/blob/f52cc3893b2797c7e25761fc24f7ed2e6859c561/src/main/java/forestry/Forestry.java#L71

Look through the [forge changelog here](http://files.minecraftforge.net/maven/net/minecraftforge/forge/1.12.2-14.23.3.2676/forge-1.12.2-14.23.3.2676-changelog.txt) to find which version adds the oredict entries, and use that version as the minimum required.

mezz (changes requested):

temp1011:
```
Build 1.12.2-14.23.2.2643:
	34900092+temp1011: Add fences and fence gates to ore dictionary (#4807)
```

I believe this should be the right version

mezz (approved):
Looks good to me.
@Nedelosk if you can test this, I think it will be ready to merge.

Nedelosk:
It would be good if you could add more information to the comments of the config options. Things like how an entry has to look like. `Added / Blacklisted items` is not much information for a user that sees the config for the first time.

Nedelosk (changes requested):
.

Nedelosk:
@temp1011 Have you tested it yourself ? You should always test thing if you make a PR. You are parsing the config at `doInit` but all creates get created in the `RegistryEvent.Register<Item>` event and that is called directly after `preInit`. So currently you can only add items with the config.

src/main/java/forestry/storage/ModuleCrates.java
```diff
@@ -37,6 +45,11 @@
 @ForestryModule(moduleID = ForestryModuleUids.CRATE, containerID = Constants.MOD_ID, name = "Crate", author = "SirSengir", url = Constants.URL, unlocalizedDescription = "for.module.crates.description")
 public class ModuleCrates extends BlankForestryModule {
 
+	private static final String CONFIG_CATEGORY = "crates";
+
+	public static final List<String> cratesRejectedOreDict = new ArrayList<>();
+	public static List<ItemStack> cratesRejectedItem = new ArrayList<>();
```
Nedelosk: Can you please change this to a map with an `Item` key and and a `List` value or you use a `Multimap`. This way you not have to iterate over every stack even it the stack has nothing todo with the item.

src/main/java/forestry/storage/CrateRegistry.java
```diff
@@ -61,6 +61,9 @@ private static void registerCrate(ItemStack stack, @Nullable String oreDictName)
 
 	@Override
 	public void registerCrate(String oreDictName) {
+		if (ModuleCrates.cratesRejectedOreDict.contains(oreDictName)) {
```
Nedelosk, line 65: Can you please add a `blacklistCrate` function to the 'ICrateRegistry' interface.

src/main/resources/assets/forestry/lang/en_us.lang
```diff
@@ -1859,6 +1859,11 @@ for.config.butterfly.entities.pollination.comment=Allow butterflies to pollinate
 for.config.backpacks.item.stacks.format=Add itemStacks for the %s's backpack here in the format 'modid:name:meta'. For wildcard metadata the format is 'modid:name'.
 for.config.backpacks.ore.dict.format=Add ore dictionary names for the %s's backpack here in the format 'oreDictName'. Regular expressions may also be used.
 
+for.config.crates.items.accepted.comment=Added items
```
Nedelosk: It would be good if you could add more information to the comments of the config options. Things like how an entry has to look like or a list of all entries that can be blacklisted. `Added / Blacklisted items` is not much information for a user that sees the config for the first time.

src/main/java/forestry/storage/ModuleCrates.java
```diff
@@ -67,6 +80,105 @@ public void preInit() {
 		MinecraftForge.EVENT_BUS.register(this);
 	}
 
+	@Override
+	public void doInit() {
```
Nedelosk: You are parsing the config at `doInit` but all creates are created in the `RegistryEvent.Register<Item>` event which gets called directly after `preInit`.

Nedelosk (changes requested):
Have you tested it yourself ? You should always test thing if you make a PR. Currently you can only add items with the config.

temp1011:
I thought I had it working... I haven't addressed everything yet because I don't have enough time. However, I do wonder whether having this as a Json might be nice. But I will fix this first and then think about that when working on the Json recipes.

temp1011:
If this needs better support let me know.

Nedelosk (approved):

PR #2050, opened 2018-03-23
Move null checks for registering crates and backpacks

temp1011:
I'm splitting up https://github.com/ForestryMC/ForestryMC/pull/2024 to make it easier to review (I hope). This PR doesn't add much but I think it's cleaner to check for `null` objects inside the methods rather than before registering them, at least for plugins. It also reduces the chances of a crash because an item/block is null for some reason. I could add a log warning if we're worried about things silently failing.

### DIFF
--- a/src/main/java/forestry/api/storage/IBackpackInterface.java
+++ b/src/main/java/forestry/api/storage/IBackpackInterface.java
@@ -11,9 +11,10 @@ import java.util.function.Predicate;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 
-import net.minecraftforge.fml.common.registry.GameRegistry;
 import net.minecraftforge.oredict.OreDictionary;
 import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import net.minecraftforge.fml.common.registry.GameRegistry;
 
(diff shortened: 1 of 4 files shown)


src/main/java/forestry/api/storage/IBackpackInterface.java
```diff
@@ -39,7 +40,7 @@
 	 * @param itemStack   The itemStack that the backpack should accept.
 	 *                    {@link OreDictionary#WILDCARD_VALUE} can be used for meta value.
 	 */
-	void addItemToForestryBackpack(String backpackUid, ItemStack itemStack);
+	void addItemToForestryBackpack(String backpackUid, @Nullable ItemStack itemStack);
```
Nedelosk, line 43: @mezz Do you think allowing `null` for the api methods is a good idea ?

temp1011:
This is solved by `ModuleHelper`

src/main/java/forestry/api/storage/IBackpackInterface.java
mezz: `ItemStack` should almost never be null, it has to be a very special case.

PR #2053, opened 2018-03-24
Clean up code

temp1011:
This is the stuff that analysis bought up that I couldn't find any reason to justify keeping. It's mostly just minor stuff like imports. There's probably a lot more that could be done but I'm trying not to break anything in the process and not have a diff going into hundreds of lines.


temp1011:
https://github.com/ForestryMC/ForestryMC/blob/54a9da917dc09cc6e7800001cd1eee7320021da1/src/main/java/forestry/greenhouse/multiblock/blocks/world/NextTickEntry.java#L38

~This was so weird I feel like there is actually a reason for this (probably some concurrency stuff over my head).~

src/main/java/forestry/core/utils/Translator.java
```diff
@@ -2,25 +2,24 @@
...
 
-	@SuppressWarnings("deprecation")
 	public static String translateToLocal(String key) {
-		if (I18n.canTranslate(key)) {
-			return I18n.translateToLocal(key);
+		if (I18n.hasKey(key)) {
+			return I18n.format(key);
 		} else {
-			return I18n.translateToFallback(key);
+			return net.minecraft.util.text.translation.I18n.translateToFallback(key);    //no equivalent from what I can see
```
temp1011: This doesn't seem to have a replacement. If we don't use it the unformatted String will be displayed. Not sure what the best option is.

mezz:
I don't think it's very safe to make these `I18n` changes, can you separate them into a different commit?

src/main/java/forestry/arboriculture/genetics/TreeDefinition.java
```diff
@@ -973,9 +974,12 @@ protected void registerMutations() {
 		IGermlingModelProvider germlingIconProvider = ModelProviderFactory.create(woodType, uid, leafIconProvider);
 		IWoodProvider woodProvider = WoodProviderFactory.create(woodType);
 
-		IAlleleTreeSpeciesBuilder speciesBuilder = TreeManager.treeFactory.createSpecies(uid, unlocalizedName, "Sengir", unlocalizedDescription, dominant, branch.getBranch(), binomial, Constants.MOD_ID, leafIconProvider, germlingIconProvider, woodProvider, this);
+		ILeafProvider leafProvider = new LeafProvider();
+
+		IAlleleTreeSpeciesBuilder speciesBuilder = TreeManager.treeFactory.createSpecies(uid, unlocalizedName, "Sengir", unlocalizedDescription, dominant, branch.getBranch(), binomial, Constants.MOD_ID, leafIconProvider, germlingIconProvider, woodProvider, this, leafProvider);
 		setSpeciesProperties(speciesBuilder);
 		this.species = speciesBuilder.build();
+		leafProvider.init(this.species);
```
Nedelosk: You don't have to call `init` manually it gets called in the `build` method of the species. https://github.com/ForestryMC/ForestryMC/blob/36ccd5c3e1f1aa3f600bbcdd6111b2bd090c64ca/src/main/java/forestry/arboriculture/genetics/alleles/AlleleTreeSpecies.java#L87-L92

Nedelosk (changes requested):

temp1011:
`IL8N` removed, conflicts fixed.

src/main/java/forestry/arboriculture/models/SimpleModel.java
```diff
@@ -154,7 +152,7 @@ public IModel retexture(ImmutableMap<String, String> textures) {
 				try {
 					model = ModelLoaderRegistry.getModel(loc);
 				} catch (Exception e) {
-					throw Throwables.propagate(e);
+					throw new RuntimeException(e);
```
mezz: this should be
```java
Throwables.throwIfUnchecked(e);
throw new RuntimeException(e);
```
That way the exception is only wrapped if it needs to be.

You can find out more about it here: https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

mezz (changes requested):

temp1011:
Pretty sure the reflection for `CriteriaTriggers.register()` isn't needed any more.

mezz (approved):

Nedelosk (approved):

Issue #2060, opened 2018-03-25
Hives causing performance issues


Sir-Will:
hey,
any chance this can be improved?

forestry_1.12.2-5.8.0.250

![](https://i.imgur.com/rPkfVZF.jpg)

temp1011:
If this can be moved without causing issues probably because then most of that won't happen every tick. However, I'm sure mezz had good reasons for keeping them outside.

https://github.com/ForestryMC/ForestryMC/blob/54a9da917dc09cc6e7800001cd1eee7320021da1/src/main/java/forestry/apiculture/WorldgenBeekeepingLogic.java#L95-L96

temp1011:
Also, was this while generating new chunks?

Sir-Will:
No, world was pre-generated.

temp1011:
This should be improved now

temp1011:
Hey, did you ever re-run the profiler? Has this actually helped or are hives still causing performance issues for you? Just curious to see if there was a good impact.

Sir-Will:
Unfortunately there was no modpack update yet. I have to wait for the modpack to update forestry.

temp1011:
Has it updated now?

Sir-Will:
Just checked and I couldn't find it in the snapshot at all. I assume your fix improved it a lot, thanks.

temp1011:
No problem

PR #2063, opened 2018-03-27
Optimise Worldgen Hives

temp1011:
Possibly I'm misunderstanding something here. However, I think some things can be moved in `WorldgenBeekeepingLogic` to further optimise worldgen hives.

- As far as I'm aware the queen in these hives doesn't change so it can be stored as a field
- I don't think the flower cache needs to be updated if we're not checking whether the bee can work or not. So it can be updated less often.

Could help for https://github.com/ForestryMC/ForestryMC/issues/2060.

### DIFF
--- a/src/main/java/forestry/apiculture/BeekeepingLogic.java
+++ b/src/main/java/forestry/apiculture/BeekeepingLogic.java
@@ -56,6 +56,7 @@ import forestry.core.config.Constants;
 import forestry.core.errors.EnumErrorCode;
 import forestry.core.utils.Log;
 import forestry.core.utils.NetworkUtil;
+import forestry.core.utils.TickHelper;
 
 public class BeekeepingLogic implements IBeekeepingLogic {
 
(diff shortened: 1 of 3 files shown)


mezz:
The flower cache is only used to implement `WorldgenBeekeepingLogic#getFlowerPositions` so that the bee particles fly to nearby flowers. You can update it less often but there are some things to consider. Internally it uses `HasFlowersCache#FLOWER_CHECK_INTERVAL` for updating every 200 ticks. If you make your change here, it will only update once every 33 minutes (200 * 200 ticks). That seems like a long time, maybe it's fine though.

temp1011:
33 minutes is probably too long. 20 seconds sounds better.

mezz:
Ok now apiaries are checking every 2 ticks... `HasFlowersCache` is shared by everything that houses bees.

temp1011:
I misread your comment as the flower cache is only implemented by `WorldgenBeekeepingLogic#getFlowerPositions`. Woops. Slowing down `BeekeepingLogic` seems a bit ugly but I have to keep to the lowest common update interval I think. Could be problematic if addons use this code.

mezz:
How about pass an interval in to the flower cache as a parameter in its constructor?

temp1011:
Yes, that would be a sensible solution that I didn't think of :smile:

Issue #2062, opened 2018-03-27
[1.12.2] Bottler: Emptied cans are deleted


Syndesi:
Hi,

if I put any filled tin can into the Bottler, than the fluid is put into the internal tank but the empty can is deleted, therefore a closed system isn't possible.
Is this a wanted feature? And if so, could you please add a config option so that the cans aren't destroyed anymore?

temp1011:
I think this can definitely be a config option for capsules themselves.

Syndesi:
Woah, thank you very much for your fast response :D

temp1011:
No problem.

PR #2066, opened 2018-03-27
Add config to let capsules be reusable

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2062. I think this is valid option given that capsules picking up stuff in world also is.


mezz (approved):
Looks good to me.
I don't think it's very balanced but I'm fine with having it as a config.

Syndesi:
Very cool, thank you all :D

PR #2068, opened 2018-03-29
Start removing metadata from code in preparation for 1.13

temp1011:
Seems that metadata is changing in 1.13 to not used by mods. I'm hoping to move away from it where possible in current code to reduce the work to do for the update.

### DIFF
--- a/src/main/java/forestry/core/network/packets/PacketFXSignal.java
+++ b/src/main/java/forestry/core/network/packets/PacketFXSignal.java
@@ -22,6 +22,8 @@ import net.minecraft.block.SoundType;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.client.Minecraft;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTUtil;
 import net.minecraft.util.SoundCategory;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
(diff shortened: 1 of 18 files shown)


src/main/java/forestry/core/network/packets/PacketFXSignal.java
```diff
@@ -80,10 +77,9 @@ public void onPacketData(PacketBufferForestry data, EntityPlayer player) throws
 			BlockPos pos = data.readBlockPos();
 			VisualFXType visualFX = VisualFXType.values()[data.readByte()];
 			SoundFXType soundFX = SoundFXType.values()[data.readByte()];
-			Block block = Block.getBlockById(data.readVarInt());
-			IBlockState blockState = block.getStateFromMeta(data.readVarInt());
-
 			World world = player.world;
+			IBlockState blockState = world.getBlockState(pos);
```
temp1011: I'm not sure how bad this is for performance but I would say that looking for an ID in the block registry is probably better than searching the world for a block at a specific position.

Nedelosk: Why have you changed it if the other way was better ?

Nedelosk:
~I don't think it makes sense to move away from the metadata in 1.12.2. In many cases we anyway have to create a new block instead of using metadata. @mezz What do you think ?~
Ignore this comment. In the cases of this PR it should be ok.

src/main/java/forestry/core/network/packets/PacketFXSignal.java
mezz: The performance difference here is not a concern, but there is a possible race condition when we place/break a block and immediately fire some FX packets. The client may not have updated the world to match the server when it gets this packet, so it could get the wrong blockState here.

mezz (changes requested):

temp1011:
`PacketFXSignal` now just writes and reads the block from NBT. This uses properties so should be reasonably update safe.

mezz (approved):

Issue #2069, opened 2018-03-29
Recent crash in 1.12.2,


CarlosMZ12268:
https://gist.github.com/Carlosedu1/fd4802d9898ee880442f294e86dd0b5d

Recent crash I had on 1.12.2, when troubleshooting a few mods. It mentions Forestry several times, which led me to seek help here. Tried it on both recommended and latest forge versions, and all other mods listed in the log are in their latest versions.

temp1011:
~I think the only one of those mods we add a fermenter recipe for is biomes o plenty. Does the crash still happen if you remove that/disable the plugin for it?
Not BOP. Better with mods and better with addons both have newer versions available on curseforge from what I can see. Remove stuff which *shouldn't* be using the fermenter (from what I would expect) like aroma's stuff, bibliocraft, architecturecraft etc. then remove one by one to find the mod. From trying a few things I suspect it's better with addons but I couldn't get the game to launch with it installed (different crash).~
Ignore me

Nedelosk:
@temp1011 Forestry crashes at the factory module and not at the module of another mod.

> 	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:122)
> 	at forestry.factory.recipes.FermenterRecipe.<init>(FermenterRecipe.java:35)
> 	at forestry.factory.recipes.FermenterRecipeManager.addRecipe(FermenterRecipeManager.java:37)
> 	at forestry.core.recipes.RecipeUtil.addFermenterRecipes(RecipeUtil.java:49)
> 	at **forestry.factory.ModuleFactory.registerRecipes(ModuleFactory.java:401)**

temp1011:
Hmm well I just successfully launched with better with mods and better with addons so it's not them. I don't think any of the other mods should be a problem though.

temp1011:
Found while looking at an unrelated log: 
```
[17:35:43] [main/INFO]: Potentially Dangerous alternative prefix `minecraft` for name `brown_mushroom`, expected `betterwithmods`. This could be a intended override, but in most cases indicates a broken mod.
[17:35:43] [main/INFO]: Potentially Dangerous alternative prefix `minecraft` for name `red_mushroom`, expected `betterwithmods`. This could be a intended override, but in most cases indicates a broken mod.
```
So I guess mushrooms are being overriden. I'm not sure how much/if things break though.

CarlosMZ12268:
[...]

PR #2071, opened 2018-03-30
Possible fix fermenter crash

temp1011:
Looks like the crash happened on the line which registers the mushroom fermenter recipe in https://github.com/ForestryMC/ForestryMC/issues/2069. This may resolve it or it may not. I suspect it could be this though since some mods seem to override standard mushroom behavior. I think mushrooms should get an ore dictionary entry though, lots of mods add different kinds and it seems silly that we can't just register one recipe for all of them.

### DIFF
--- a/src/main/java/forestry/factory/ModuleFactory.java
+++ b/src/main/java/forestry/factory/ModuleFactory.java
@@ -58,6 +58,7 @@ import forestry.core.items.ItemRegistryCore;
 import forestry.core.items.ItemRegistryFluids;
 import forestry.core.network.IPacketRegistry;
 import forestry.core.recipes.RecipeUtil;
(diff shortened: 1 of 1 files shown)


mezz:
I don't think this will resolve the issue.
We should try to reproduce this issue before fixing it.

temp1011:
I'll try again and see if I can get better with mods and better with addons to work by themselves.

temp1011:
~~~https://gist.github.com/temp1011/5e8c9edddd214bbebf7d3eae0ca2c8a2. Looks like it's a coremod which will make fixing things difficult (I don't see any other reason a mod would try to find an obsfucated field which is crashing the dev instance).~~~
Maybe not, but something is wrong. It shouldn't crash there yet does. I have no idea how coremods work but that seems suspicious.

PR #2074, opened 2018-04-05
Update EnumTemperature.java

temp1011:
See Discord for the crash


temp1011:
~I'm not quite sure how this has happened. `getDefaultTemperature()` shows in my forge workspace but not forestry. Maybe we need to change forge version? Maybe I'm wrong.~
Just need forge update to fix mappings hopefully.

temp1011:
Broken.

PR #2058, opened 2018-03-24
Plugin tech reborn

temp1011:
Requires https://github.com/ForestryMC/ForestryMC/pull/2050.

Closes https://github.com/ForestryMC/ForestryMC/issues/2019.

Adds support for https://github.com/TechReborn/TechReborn.

PR #2078, opened 2018-04-09
Tech reborn plugin

temp1011:
https://github.com/ForestryMC/ForestryMC/pull/2058 messed up. Now uses module helper.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,10 @@ repositories {
 		name = "ic2"
 		url = "http://maven.ic2.player.to/"
 	}
+	maven {
+		name = "Modmuss50"
+		url = "http://maven.modmuss50.me"
+	}
 }
 
 dependencies {
@@ -66,6 +70,7 @@ dependencies {
 	deobfCompile "mezz.jei:jei_${jei_mcversion}:${jei_version}:api"
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
+	compile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
 }
 
 configurations {
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
 	deobfCompile "mezz.jei:jei_${jei_mcversion}:${jei_version}:api"
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
-	compile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
+	deobfCompile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
 }
 
 configurations {
--- a/src/main/java/forestry/farming/logic/crops/CropRubber.java
+++ b/src/main/java/forestry/farming/logic/crops/CropRubber.java
@@ -45,12 +45,12 @@ public class CropRubber extends CropDestroy {
 	private static <T extends Comparable<T>> IBlockState getReplantState(IBlockState sappyState) {
 		if (hasRubberToHarvest(sappyState)) {
 			for (Map.Entry<IProperty<?>, Comparable<?>> wetPropertyEntry : sappyState.getProperties().entrySet()) {
-				if (wetPropertyEntry.getKey() instanceof PropertyBool && wetPropertyEntry.getKey().getName().equals("hassap")) {
-					return sappyState.withProperty(PropertyBool.create("hassap"), false);
-				}
 				String valueWetString = wetPropertyEntry.getValue().toString();
 				String valueDryString = valueWetString.replace("wet", "dry");
 				IProperty<?> property = wetPropertyEntry.getKey();
+				if (property instanceof PropertyBool && property.getName().equals("hassap")) {
+					return sappyState.withProperty(PropertyBool.create("hassap"), false);
+				}
 
 				IBlockState baseState = sappyState.getBlock().getBlockState().getBaseState();
 				IBlockState dryState = getStateWithValue(baseState, property, valueDryString);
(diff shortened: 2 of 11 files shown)


temp1011:
If metadata is completely gone and there isn't a good way to get blockstates through forge it may be worth to start using `@Opitional.Interface` in forge. That way we don't have to include the interface but still get access to it. The disadvantage is forestry then has to load after that mod.

Nedelosk:
@temp1011 What are you talking about ?

temp1011:
https://github.com/MinecraftForge/MinecraftForge/blob/d0bdb233000ac9aff049c5ae2ad8043480a96af9/src/main/java/net/minecraftforge/fml/common/Optional.java.
So Forestry can use eg - the API for Tech Reborn but doesn't have to include it. Maybe this can be done anyway though.

temp1011:
@Nedelosk I've changed some things to use the tech reborn interface in the plugin. This should run fine without Tech Reborn installed. However, for the dev environment I think the API will be needed so things compile. Luckily we can get this off of the curse maven API.

src/main/java/forestry/farming/logic/crops/CropRubber.java
```diff
@@ -43,6 +45,9 @@ public CropRubber(World world, IBlockState blockState, BlockPos position) {
 	private static <T extends Comparable<T>> IBlockState getReplantState(IBlockState sappyState) {
 		if (hasRubberToHarvest(sappyState)) {
 			for (Map.Entry<IProperty<?>, Comparable<?>> wetPropertyEntry : sappyState.getProperties().entrySet()) {
+				if (wetPropertyEntry.getKey() instanceof PropertyBool && wetPropertyEntry.getKey().getName().equals("hassap")) {
```
mezz: Below this we have `IProperty<?> property = wetPropertyEntry.getKey();`, you should use `property` here instead of calling `getKey()` more times.

src/main/java/forestry/farming/logic/crops/CropRubber.java
```diff
@@ -60,19 +65,22 @@ public CropRubber(World world, IBlockState blockState, BlockPos position) {
...
+		} else if (PluginTechReborn.RUBBER_WOOD != null && ItemStackUtil.equals(block, PluginTechReborn.RUBBER_WOOD)) {
+			return blockState.getValue(PropertyBool.create("hassap"));
 		}
 		return false;
 	}
 
 	@Nullable
-	private static <T extends Comparable<T>> IBlockState getStateWithValue(IBlockState baseState, IProperty<T> property, String valueString) {
+	private static <T extends Comparable<T>> IBlockState getStateWithValue(IBlockState
+			baseState, IProperty<T> property, String valueString) {
```
mezz: I'd prefer to keep this on one line

src/main/java/forestry/farming/logic/crops/CropRubber.java
```diff
@@ -83,8 +91,11 @@ public static boolean hasRubberToHarvest(IBlockState blockState) {
 	@Override
 	protected NonNullList<ItemStack> harvestBlock(World world, BlockPos pos) {
 		NonNullList<ItemStack> harvested = NonNullList.create();
-		harvested.add(PluginIC2.resin.copy());
-
+		if (PluginIC2.rubberWood != null && ItemStackUtil.equals(world.getBlockState(pos).getBlock(), PluginIC2.rubberWood)) {
```
mezz: Store `world.getBlockState(pos).getBlock()` in a variable so it is never called twice.

src/main/java/forestry/plugins/PluginTechReborn.java
```diff
@@ -0,0 +1,136 @@
...
+
+	@Override
+	@Optional.Method(modid = "techreborn")
+	public void registerRecipes() {
+		RecipeManagers.fabricatorManager.addRecipe(ItemStack.EMPTY,
+				Fluids.GLASS.getFluid(500),
+				ModuleCore.items.tubes.get(EnumElectronTube.RUBBER, 4),
+				new Object[]{" X ", "#X#", "XXX", '#', "dustRedstone", 'X', "itemRubber"});
+
+		ItemRegistryApiculture beeItems = ModuleApiculture.getItems();
```
mezz: Check that `ModuleApiculture` is loaded, this line will crash when the apiculture module is disabled in the config.

mezz (changes requested):

temp1011:
I've added the maven to `build.gradle` because I believe it's necessary for compilation in a dev environment. Is the `@Optional.Interface` stuff OK?

build.gradle
```diff
@@ -59,13 +59,18 @@ repositories {
...
+		url = "http://maven.modmuss50.me"
+	}
 }
 
 dependencies {
 	deobfCompile "net.darkhax.tesla:Tesla-${tesla_mcversion}:${tesla_version}"
 	deobfCompile "mezz.jei:jei_${jei_mcversion}:${jei_version}:api"
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
+	compile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
```
mezz: Should this be `deobfCompile` like the others?

src/main/java/forestry/plugins/PluginTechReborn.java
```diff
@@ -0,0 +1,138 @@
...
+
+	@Override
+	public void preInit() {
+		IFarmProperties rubberFarm = FarmRegistry.getInstance().registerLogic("farmRubber", FarmLogicRubber::new);
+
+		Circuits.farmRubberManual = new CircuitFarmLogic("manualRubber", rubberFarm, true);
+	}
+
+	@Override
+	@Optional.Method(modid = "techreborn")
```
mezz: Please use `MOD_ID` instead of `"techreborn"` for these

mezz (changes requested):
Looking good, `@Optional` is fine for this.

mezz (approved):

temp1011:
The Techreborn API does give me some errors when running minecraft in dev about incompatible classloaders which 'will not work'. Not sure why I get those but I have similar errors for JEI and greenhouse so it doesn't seem to cause any issues.

PR #2079, opened 2018-04-09
add invtweaks to build.gradle

temp1011:
Inventory tweaks was causing problems for me. Since curse has a maven API I thought it would be worthwhile to add it to the developer project to make setting up a workspace easier.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
 	deobfCompile "mezz.jei:jei_${jei_mcversion}:${jei_version}:api"
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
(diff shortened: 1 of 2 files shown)


temp1011:
It's easier to access the other maven and see what's there, curseforge requires an API token to check theirs through a web browser.

mezz:
I don't really like using invtweaks, I would prefer it was not part of the developer workspace by default.
It uses a coremod, which is probably why it is difficult to get working. Ideally we should not test Forestry in an environment with coremods, since they can change the game behavior in unexpected ways.

temp1011:
I'll see if I can fix the issues I was having. The API doesnt contain any coremod parts though from what I can see.

temp1011:
~So IDEA seems to forget where everything is when switching branches for me. Not sure why but it's easy to fix every time but slightly annoying.~ It just can't find the API folder. Maybe something to do with linux? It's not a huge problem to fix either way.

PR #2104, opened 2018-04-26
move lang to main project

temp1011:
This PR moves the lang submodule into the main project as discussed on discord. It will definitely make implementing new localisations in parallel with new features a lot easier. I followed this guide: https://stackoverflow.com/questions/1759587/un-submodule-a-git-submodule. If we need to preserve the history of the module then it will be a bit more complex.

If this is merged then https://github.com/ForestryMC/Forestry-Localization will need to be archived/made clear that it is no longer used. There don't seem to be any significant non stale issues or PRs open at the moment so I think this is a reasonable time to move. Localisation is changing in 1.13 anyway so we could move lang over when that branch is created.


temp1011:
I guess lang could be under resources?

mezz:
The standard location for lang is `resources/assets/<modid>/lang`. To keep things standard and simple I think they should go there.

temp1011:
I believe that `lang` is now in the default directory for lang files so should work by default now. Maybe there's something wrong though.

Nedelosk:
You have to move the files into `src/main/resources/assets/forestry/lang` and not into `src/main/resources/assets/forestry/lang/assets/forestry/lang/`.

temp1011:
It would make sense not to duplicate the path :smile:

Nedelosk (approved):

.gitignore
```diff
@@ -9,7 +9,6 @@
 
 # files
 !/.gitignore
-!/.gitmodules
```
mezz, line 12: I don't think this is supposed to be ignored

src/main/resources/assets/forestry/lang/cs_CZ.lang
```diff
@@ -0,0 +1,914 @@
+####################
+# CORE
+####################
+
+yes=Yes
```
mezz: It looks like these keys are extremely outdated and will not work at all.
Please do a quick look and remove any lang files that have never been updated since like 1.6.4, they just add size to the jar. Anyone who wants to translate them in the future can look at the old repo as a starting point.

mezz (changes requested):

temp1011:
I've removed everything that hasn't been updated since 2014. Hopefully that's about the right amount of time you meant. I'm not sure why the manual localisation part isn't automatically added. I'd like to move that to this project too.

temp1011:
Manual stuff added.

mezz:
can you fix or explain the gitignore?

mezz:
Oh I was reading it backwards. I wonder why it was ignored before.

mezz (approved):

mezz:
I think `LANG.md` should go in the lang folder too

Nedelosk:
@temp1011 You have used the wrong branch, the most lang files are from 1.7.10. Thats why the manual localisation part wasn't automatically added.

temp1011:
Sorry my fault. I think that I must have used `master` instead because I mistakenly thought that it would be on the latest version.

Hopefully nothing else is broken. Maybe one day I'll submit something that doesn't cause a problem.

mezz:
No big deal. We make less mistakes over time but it's always going to happen, especially starting out.

Issue #2090, opened 2018-04-18
Crash when using can (capsule) 


Dragon905:
[crash-2018-04-18_03.39.16-client.txt](https://github.com/ForestryMC/ForestryMC/files/1922675/crash-2018-04-18_03.39.16-client.txt)

Crashes when using can (capsule) on lava. Haven't tried with other liquids yet.
I also set the "capsules can pick up fluids from the world" to true in the config.

Dragon905:
I forgot to mention that it happens with random quantities of source blocks. For example: when I pick up 3 lava source blocks it crashes, or when I pickup 50 lava source blocks.

PR #2105, opened 2018-04-26
fix crash when raytrace result is null and using can on liquid in world

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2090.


temp1011:
It looks like there are more places where this might happen too, see https://github.com/MC4EP/ForestryMC/commit/38c5338f8bfb8115a65eaa70c2d7d8db4c757cbe.

PR #2109, opened 2018-04-27
remove deprecated I18n

temp1011:
Replace deprecated methods with current ones. Unfortunately this means that lang files without a key will just return an unformatted string in `Translator`. This is solvable by making every lang file include the translations for `en_US` by default if no key exists. This also makes translation a bit easier IMO. This should be relatively easy to set up through a script. Ideally I would wait for https://github.com/ForestryMC/ForestryMC/pull/2104 though.

This is split off from https://github.com/ForestryMC/ForestryMC/pull/2053.


mezz (changes requested):
Make sure to test this on a server: [crash log](https://gist.github.com/mezz/2750107c533bdc58273ac3f440af644e)

temp1011:
Servers will no longer crash. However, there's probably a better way to do this as my knowledge of sidedness isn't great. Would it be better if I didn't move `Translator` to the API and instead add `ITranslator`? Is there a better way to do this? (it seems that a lot of these methods will only be called client side anyway).

Nedelosk:
You should take a look at the translator of binnie.
https://github.com/ForestryMC/Binnie/blob/master-MC1.12/core/src/main/java/binnie/core/util/I18N.java
https://github.com/ForestryMC/Binnie/blob/master-MC1.12/core/src/main/java/binnie/core/proxy/I18NServer.java
https://github.com/ForestryMC/Binnie/blob/master-MC1.12/core/src/main/java/binnie/core/proxy/I18NClient.java
https://github.com/ForestryMC/Binnie/blob/master-MC1.12/core/src/main/java/binnie/core/proxy/I18NProxy.java

temp1011:
Thank you the proxy makes things pretty easy. I'm 99% sure everything in the API involving translation should only ever be called client side so doesn't need to be changed. I think the main problem is that translation doesn't really work server side anymore so we should just change the code to make sure nothing is called that shouldn't be.

Nedelosk:
@mezz Could you please review this again ?

temp1011:
It's probably best to just wait for 1.13 here.

PR #2111, opened 2018-04-28
JSON recipes

temp1011:
This is no where near complete but I'm opening this now for review/feedback purposes. It might not be possible to merge until 1.13 anyway.

This PR adds (or begins to add) JSON recipes for crafting recipes. If this goes through then I will move on to machine recipes.

I've used a JSON recipe converter found at https://gist.github.com/williewillus/a1a899ce5b0f0ba099078d46ae3dae6e. Hopefully the license there is enough for licensing things. If not I'm sure we could ask the owner of it.

Currently the game will only launch with **all** modules except apiculture disabled because I haven't adapted those yet since there are still bugs to iron out. In particular, at the moment recipe names aren't great and recipe overlap on things like combs seems to break the recipe (maybe there's a way to have all the recipes in one file?).

Recipes won't load if the module that loads them is disabled. The performance impact of this check might be concerning but I haven't checked it yet. If it is we can just store the recipes by module in folders and use a custom parser to ignore the folder if the module is disabled.

Having recipes stored like this should make changing/deleting them much easier. There's a config option to reset recipes if needed.

I have no idea how this interacts with existing worlds so it probably needs to wait for 1.13 or maybe I can write a converter.

TLDR: Sorry for wall of text, JSON recipes, not working perfectly yet.

Feedback/comments appreciated.


src/main/java/forestry/core/recipes/RecipeUtil.java
```diff
@@ -132,28 +131,24 @@ private static ItemStack getCraftingEquivalent(NonNullList<ItemStack> stockCopy,
...
 	}
 
-	public static void addRecipe(String recipeName, Item item, Object... obj) {
-		addRecipe(recipeName, new ItemStack(item), obj);
+	public static void addRecipe(String moduleUID, Item item, Object... obj) {
+		addRecipe(moduleUID, new ItemStack(item), obj);
 	}
 
-	public static void addRecipe(String recipeName, ItemStack itemstack, Object... obj) {
-		ShapedRecipeCustom recipe = new ShapedRecipeCustom(itemstack, obj);
```
Nedelosk, line 152: Are you sure that `ShapedRecipeCustom` is no longer needed ?

src/main/resources/assets/forestry/recipes/bee_combs_0_16.json
```diff
@@ -0,0 +1,24 @@
+{
+  "result": {
+    "item": "forestry:bee_combs_0",
+    "data": 16
```
Nedelosk: Seems like there's a bug in my recipe generation system. A block can only have 15 metas.

src/main/java/forestry/core/recipes/RecipeUtil.java
temp1011: Hmm, I don't think the current JSONs allow moving smaller recipes in the crafting grid (although they do use oredict). I'm hopeful there might be a better way to do this in the file rather than register one recipe for each position in the crafting grid though. It might require a custom loader.

Nedelosk: Allow moving smaller recipes in the crafting grid works with vanilla recipes as well. The sense behind `ShapedRecipeCustom` is another one. In 1.7.10 the recipe never checked the nbt of the ingredients so we had to add a custom recipe for the farm.

Nedelosk: I atleast think that that is the sense behind it. @mezz please correct me if I am wrong.

src/main/resources/assets/forestry/recipes/bee_combs_0_16.json
temp1011: https://github.com/ForestryMC/ForestryMC/pull/2112

src/main/java/forestry/core/recipes/RecipeUtil.java
temp1011: Ah OK I see what you mean. I had a look and it seems you can get around this. It looks like CoFH also has recipes involving NBT tags (liquids in tanks and owners on devices) and it preserves them like this:
https://github.com/CoFH/CoFHCore/blob/dev/src/main/java/cofh/core/util/crafting/ShapedUpgradeRecipeFactory.java
https://github.com/CoFH/ThermalExpansion/blob/dev/src/main/resources/assets/thermalexpansion/recipes/_factories.json
https://github.com/CoFH/ThermalExpansion/blob/dev/src/main/resources/assets/thermalexpansion/recipes/satchel_1.json
I just need to implement that.

Nedelosk: There should be a IngredientNBT.

temp1011: Seems like it should be. Now I have to just understand how the system works properly. Would have helped if there was better documentation...

temp1011: Seems like NBT is already built in ?
https://github.com/temp1011/ForestryMC/blob/3083abad1ef24568ee840c802bf992a3ce1f7a0c/src/main/java/forestry/core/recipes/json/RecipeConverter.java#L189-L192

temp1011:
### TODO:
- reduce duplication in files - implement something in `_constants` that iterates over metadata (or variant property ideally for future proofing so similar recipes can be defined in one file).
- make names better on files
- add machine JSON recipes
- check all recipes work (NBT etc)
- `database_0_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt.json`
- `greenhouse.window_up_0_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt_alt.json`

Storing things as JSON data could be very powerful. The only problem is the size of files and bulk. Hopefully with a bit of adaptation that could be reduced. I think it would be interesting to consider storing other date eg - mutations in JSON as well.

temp1011:
Recipes that have multiple ingredient sets to craft the same thing can be deduplicated quite easily. However the largest repeat problems can't be removed in this way. Things like doors and fences need a one to one recipe mapping but for an incredibly similar structure. I should be able to make a custom implementation of `IRecipe` for these but methods like `getRecipeOutput()` become tricky (eg - `RecipeFireworks` in vanilla). I'm not sure how well the recipes would work with JEI and others.

Ideally it would be very nice if there was a `IRecipeMultiFactory` or similar that could return multiple `IRecipe` in forge. However, from what I have heard it's unlikely that something like this would get implemented because forge is aiming for explicit data in recipes.

Is it possible to work around this in forge somehow? Or should I just put Jsons for item groups that produce 20+ recipes somewhere else and automatically generate Jsons in `/recipes` (this makes customisation easy). 

Sorry for wall of text

temp1011:
Folders definitely help make this a bit cleaner. Hopefully they will work with datapacks come 1.13. I'm still working on custom recipes/something to reduce recipe duplication. Hopefully forge might implement something so datapacks can override anything you put in `resources` (that might be how they work anyway). That would mean that machine recipes could also be overriden.
EDIT: I believe this already does exist.

temp1011:
`RecipeConverter` and registering recipes using `RecipeUtil` can be deleted once everything is finished I think because there won't be any use for them.

temp1011:
@mezz sorry for ping but you have more experience with this than me. I would like to implement a custom `IRecipe` for things like turning logs into planks because then rather than having 20+ files there can just be 1. However, I am worried that such an implementation might break JEI (and possibly other things to). In this recipe, `getRecipeOutput()` would be dependent on other methods in the class being called first and similarly `getIngredients()` would not work well. It looks like vanilla does use stuff similar to this in things like `RecipeFireworks` but from what I understand it doesn't play well with JEI.

Is implementing things in this way the wrong way to go? Is there a better way to do this? I could store some recipes in another directory in `assets/` and parse them separately but that's quite ugly. Equally, having 6 recipes for each `woodType` increases the download size and makes customisation/maintenance a lot harder (the diff for this PR is much too large already IMO).

mezz:
For JEI you have two options:
1. Give JEI multiple wrappers directly using `IModRegistry#addRecipes` instead of giving it one IRecipe and a handler.
2. Write a dynamic wrapper that has all the inputs and outputs, but changes what is displayed based on the current focus from the user.

For something like these, I would do option 1. Binnie's has some crazy stuff where I would pick option 2 because it would require thousands of recipe wrappers otherwise.

temp1011:
Ok thanks, as long as I can have the recipe without breaking things I will implement it. It would be nice if there was a way to define many similar recipes in one file through forge but I suspect that if it was a good change it would have been done already.

temp1011:
Wood types are largely deduplicated now and farm blocks have had quite a few recipes removed as well. The other main source of spammy files are the greenhouse windows but given that they are likely to be replaced I'm leaving them as is right now.

The code is quite messy and JEI doesn't work at the moment but hopefully that shouldn't be awful to fix.

Recipe names + converter code I plan to fix at the end once I'm not having to change things a lot any more. 
EDIT: Removing greenhouse window would save ~1000 lines.
EDIT2: Bee comb to block recipe should probably also be a custom recipe. Having 17 is quite awkward.

temp1011:
Aside from one JEI problem all that is left for normal crafting recipes are the plugins and config options.

temp1011:
Crafting recipes are done, just need testing. Machine recipes hopefully next.

Nedelosk:
@temp1011 Have you tested this ?

temp1011:
From what I can tell this works fine. However, I noticed there are a couple of recipes missing which I can't seem to find anywhere: ash block and greenhouse screen + portable screen. Maybe I missed them when copying across?

temp1011:
I guess this should be updated based on #2184 anyway.

temp1011:
Seems that in 1.13 code recipes are gone so this may help then

temp1011:
@Nedelosk should I remove the config stuff for recipes since the player can just edit them now? For some things like gamemode it's probably best to leave the config there for convenience but for other stuff should be fine.

src/main/java/forestry/core/recipes/DisableRecipe.java
```diff
@@ -0,0 +1,49 @@
...
+		if (enabledModuleUIDs == null) {
+			getEnabledModules();
+		}
+		return () -> enabledModuleUIDs.contains(json.get("module").getAsString());
+	}
+
+	private static void getEnabledModules() {
+		enabledModuleUIDs = new HashSet<>();
+		Set<ResourceLocation> enabledModuleRLs = ForestryAPI.enabledModules;
+		enabledModuleRLs.forEach(rl -> enabledModuleUIDs.add(rl.getPath()));
```
Nedelosk: Could you please add an `"container"` option to the json so we can use this condition for binnie aswell?

Nedelosk (changes requested):

Nedelosk:
@temp1011 It seems the best to remove the config stuff with the next big update. Which probably is the version change to 1.13

temp1011:
I left the gamemode stuff in because I wasn't sure if you wanted it or not.

PR #2112, opened 2018-04-28
small fix comb recipes

temp1011:
Before when `blockCount` was 0 `blockMeta` could go up to 16. Hopefully this stops this happening and makes things slightly clearer.

### DIFF
--- a/src/main/java/forestry/apiculture/ModuleApiculture.java
+++ b/src/main/java/forestry/apiculture/ModuleApiculture.java
@@ -541,8 +541,8 @@ public class ModuleApiculture extends BlankForestryModule {
 				'B', new ItemStack(blocks.apiary),
 				'C', Items.MINECART);
 		for (int i = 0; i < EnumHoneyComb.VALUES.length; i++) {
-			int remainder = i % 16;
-			int quotient = i / 16;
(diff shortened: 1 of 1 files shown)


temp1011:
@Nedelosk hopefully this should fix the bug. Maybe I've messed up on the maths though.

src/main/java/forestry/core/recipes/RecipeUtil.java
```diff
@@ -141,6 +143,10 @@ public static void addRecipe(String recipeName, Item item, Object... obj) {
 	}
 
 	public static void addRecipe(String recipeName, ItemStack itemstack, Object... obj) {
+		if (Config.isDebug && itemstack.getMetadata() > 15) {
```
Nedelosk: Items can have metadatas higher than 15. Only blocks can't have them.

Nedelosk (changes requested):

src/main/java/forestry/core/recipes/RecipeUtil.java
temp1011: OK, I have removed this and the unnecessary commits

Nedelosk (approved):

PR #2128, opened 2018-05-07
fix bottler dupe

### LINKED ISSUE #2126: Bottler dupes liquid when draining an item that contains more than 1 bucket of fluid
Putting a fluid container that has more than 1 bucket of liquid in the bottler will fill the bottler without draining the source container, effectivly duping liquids. Tested with evilcraft tank, TE reservoir, and IE jerrycan all filled with water.  

forestry_1.12.2-5.8.0.285
Forge: 14.23.3.2655

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2126. The new behaviour will keep the stack in the emptying slot until it is completely empty but also decrease the fluid amount in it. Maybe that's not the correct way for it to work. I also ran into a weird bug where it seemed the bottler operation took longer for bigger tanks. Maybe that's intentional.

If there's a better way to fix this just ignore this one :smile: .

### DIFF
--- a/src/main/java/forestry/core/fluids/FluidHelper.java
+++ b/src/main/java/forestry/core/fluids/FluidHelper.java
@@ -74,10 +74,10 @@ public final class FluidHelper {
 		ItemStack filled = input.copy();
 		filled.setCount(1);
 
-		if(emptyStack.isEmpty()){
+		if (emptyStack.isEmpty()) {
 			emptyStack = filled;
 		}
-		
+
 		IFluidHandlerItem fluidFilledHandler = FluidUtil.getFluidHandler(filled);
 		IFluidHandlerItem fluidEmptyHandler = FluidUtil.getFluidHandler(emptyStack);
 		if (fluidFilledHandler == null || fluidEmptyHandler == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/forestry/core/fluids/FluidHelper.java
```diff
@@ -190,6 +190,9 @@ public static FillStatus drainContainers(IFluidHandler fluidHandler, IInventory
 							inv.setInventorySlotContents(outputSlot, newStack);
 							inv.decrStackSize(inputSlot, 1);
 						}
+						if(!isFillableEmptyContainer(newStack) && isFillableContainerWithRoom(newStack)) {
+							inv.setInventorySlotContents(inputSlot, newStack.copy());
```
Nedelosk: You don't have to copy the stack. It already is the copied version of `drainedItem`.

Nedelosk (changes requested):

src/main/java/forestry/core/fluids/FluidHelper.java
Nedelosk: And why do you check `isFillableContainerWithRoom` ? What if the stack only can be drained  ?

temp1011:
Ah yeah this is wrong at the moment. I'll fix it when I have time (hopefully this weekend).

temp1011:
I just need to test this but hopefully the code is better now.
EDIT: testing is proving difficult to do...

temp1011:
After much stupidity on my part I'm reasonably sure this works now. I have tested it with a thermal expansion tank and it works just fine.

Nedelosk (approved):

temp1011:
Maybe I should refactor `FluidHelper` to have base methods to check `isDrainable` and `isFillable` and tchecking for fill levels?

temp1011:
Any update on this?

mezz (approved):

PR #2134, opened 2018-05-21
Fix Tile Entity Registration names

temp1011:
Fixes errors such as 
```
21:36:47] [main/WARN] [FML]: Potentially Dangerous alternative prefix minecraft for name forestry.analyzer, expected forestry. This could be a intended override, but in most cases indicates a broken mod.
[21:36:47] [main/WARN] [FML]: Potentially Dangerous alternative prefix minecraft for name forestry.escritoire, expected forestry. This could be a intended override, but in most cases indicates a broken mod.
```
~~At the moment this will remove all tile entities in the world. So this will break things unless a datafixer is also included.~~

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 mcversion=1.12.2
-forgeversion=14.23.2.2643
+forgeversion=14.23.4.2705
 mcp_mappings=snapshot_20171003
 curse_project_id=59751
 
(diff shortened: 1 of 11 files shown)


temp1011:
I've added a datafixer. I've tested it and it seems to work for me. Given how bad the consequence of this being wrong would be it definitely needs checking again. However, this can now be applied to 1.12

Nedelosk:
@temp1011 Have you tested this ?

temp1011:
I believe I did. I will check again though.

temp1011:
It works fine for most things but the datafixer seems to cause a visual glitch (farm block appeared with wrong texture) and saplings changed to ghost blocks for some reason. It is probably safer that I remove the datafixer and this just gets merged in 1.13 (might be nice to have a branch where stuff like this can go).

Nedelosk:
https://github.com/ForestryMC/ForestryMC/commit/9b4a4c2b1d0b495f180a4dfea2bcfe492546b4d0

PR #2135, opened 2018-05-21
Include forestry prefix when logging

temp1011:
See https://gist.github.com/temp1011/291c85c53b80a9f66a418cfc7315e8ee. This makes it clearer to see which mod is doing what in big packs. Maybe there is a better way to do this through forge or logger configuration.

### DIFF
--- a/src/main/java/forestry/core/utils/Log.java
+++ b/src/main/java/forestry/core/utils/Log.java
@@ -38,7 +38,7 @@ public class Log {
 	}
 
 	private static void log(Level logLevel, String message, Object... params) {
-		LogManager.getLogger(Constants.MOD_ID).log(logLevel, String.format("[ Forestry ] %s", message), params);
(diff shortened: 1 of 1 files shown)


mezz:
No no no. Do this with logger configuration.
If you have logs like this, you're probably using the vanilla launcher which forces some bad log settings.
Try using MultiMC, it will use Forge's log settings and you will see all the proper mod prefixes.

temp1011:
Ok. I was just running from IDEA which didn't do this for some reason.

mezz:
Make sure you are using the latest Forge and IDEA, it should be working properly.

temp1011:
I'm pretty sure I am up to date. It's probably just one of a few weird errors I seem to get.
EDIT: Maybe something to do with this:
```
2018-05-27 20:58:02,635 main WARN Plugin [terminalconsole] is already mapped to class net.minecraftforge.server.console.log4j.TerminalConsoleAppender, ignoring class net.minecraftforge.server.console.TerminalConsoleAppender
```

mezz:
Hm that's pretty strange

Issue #2144, opened 2018-05-31
Bee hives can sting you when you are in the water


Leowbattle:
Most bees IRL die if they are submerged in water, so they shouldn't be able to sting you if you are underwater.

EnderiumSmith:
the ones with the flyer gene should

darthvader45:
I don't think so, since it only enables bees to work in the rain.  Working underwater would require them to be exposed to more water than a mere rainstorm.

MauveCloud:
I think it makes sense for Water bees (from Binnie's) to work underwater, since that's where their hives generate, and that might be what EnderiumSmith had in mind.  However, just checking for the tolerant flyer gene would also pick up Rocky bees, which probably shouldn't work underwater.

darthvader45:
Now that is accurate.  Would bee nice, pardon the pun.

Golden rule in Minecraft: Never dig straight down.

Nedelosk:
I am not a big fan of this can be exploited very easily.

PR #2149, opened 2018-06-05
prevent effect damage underwater

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/2144. Maybe this needs a config option?


temp1011:
I've added a config option so this behaviour is disabled by default.

PR #2158, opened 2018-06-10
BuildCraft support

dblsaiko:
Reimplements BuildCraft support. 

This includes:
 - Gate conditions
 - MJ support
 - Configuration options (power display in GUIs & toggle which power types can be used)
 - Misc stuff
 - It doesn't crash without BC

Misc stuff that doesn't work 100% (yet):
 - Has Mail trigger (doesn't work at all, no idea how to implement)
 - Engines don't output power in pulses like BC's but at a constant rate
 - Wrench doesn't work on BC machines

Compiles against BC 7.99.17, for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,9 +40,9 @@ minecraft {
 }
 
 repositories {
-	maven {
-		url 'http://maven.epoxide.xyz'
-	}
+  maven {
+    url = 'http://maven.mcmoddev.com'
+  }
 	maven {
 		url "http://dvs1.progwml6.com/files/maven"
 	}
@@ -54,14 +54,18 @@ repositories {
 		name = "Modmuss50"
 		url = "http://maven.modmuss50.me"
 	}
+  maven {
+    url = 'https://mod-buildcraft.com/maven'
+  }
 }
 
 dependencies {
-	deobfCompile "net.darkhax.tesla:Tesla-${tesla_mcversion}:${tesla_version}"
+  deobfCompile "net.darkhax.tesla:Tesla-${tesla_mcversion}:${tesla_version}"
 	deobfCompile "mezz.jei:jei_${jei_mcversion}:${jei_version}:api"
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
 	deobfCompile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
+  deobfCompile "com.mod-buildcraft:buildcraft:${buildcraft_version}"
 }
 
 configurations {
--- a/build.gradle
+++ b/build.gradle
@@ -65,7 +65,8 @@ dependencies {
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
 	deobfCompile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
-	deobfCompile "com.mod-buildcraft:buildcraft:${buildcraft_version}"
+	deobfProvided "com.mod-buildcraft:buildcraft-api:${buildcraft_version}"
+	runtimeOnly "com.mod-buildcraft:buildcraft:${buildcraft_version}"
 }
 
 configurations {
--- a/build.gradle
+++ b/build.gradle
@@ -112,9 +112,8 @@ jar {
 }
 
 task sourcesJar(type: Jar) {
-	// Not sure if this is how Abrar designed it, but grab the remapped src
+	from zipTree('build/retromapping/retromappedReplacedMain.jar') // Not sure if this is how Abrar designed it, but grab the remapped src
 	classifier = 'sources'
-	from zipTree('build/retromapping/retromappedReplacedMain.jar')
 }
 
 task apiJar(type: Jar) {
--- a/build.gradle
+++ b/build.gradle
@@ -1,16 +1,16 @@
 buildscript {
-    repositories {
-        jcenter()
-        maven { url = "http://files.minecraftforge.net/maven" }
-    }
-    dependencies {
-        classpath 'net.minecraftforge.gradle:ForgeGradle:2.3-SNAPSHOT'
-        classpath "gradle.plugin.se.bjurr.gitchangelog:git-changelog-gradle-plugin:1.47"
-    }
+	repositories {
+		jcenter()
+		maven { url = "http://files.minecraftforge.net/maven" }
+	}
+	dependencies {
+		classpath 'net.minecraftforge.gradle:ForgeGradle:2.3-SNAPSHOT'
+		classpath "gradle.plugin.se.bjurr.gitchangelog:git-changelog-gradle-plugin:1.47"
+	}
 }
 
 plugins {
-    id "com.matthewprenger.cursegradle" version "1.0.9"
+	id "com.matthewprenger.cursegradle" version "1.0.9"
 }
 apply plugin: 'net.minecraftforge.gradle.forge'
 apply plugin: 'idea'
@@ -19,7 +19,7 @@ apply plugin: "se.bjurr.gitchangelog.git-changelog-gradle-plugin"
 
 version = version_major + '.' + version_minor + '.' + version_patch + '.' + getBuildNumber()
 if (project.hasProperty('buildQualifier')) {
-    version = project.version + '-' + project.buildQualifier
+	version = project.version + '-' + project.buildQualifier
 }
 group = "net.sengir.forestry" // http://maven.apache.org/guides/mini/guide-naming-conventions.html
 archivesBaseName = "forestry_" + mcversion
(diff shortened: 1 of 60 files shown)


src/main/java/forestry/core/tiles/TilePowered.java
```diff
@@ -25,11 +24,15 @@
...
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
 
-//@Optional.Interface(iface = "buildcraft.api.tiles.IHasWork", modid = "BuildCraftAPI|tiles")
-public abstract class TilePowered extends TileBase implements IRenderableTile, ISpeedUpgradable, IStreamableGui {
+import javax.annotation.Nullable;
+import java.io.IOException;
+
+@Optional.Interface(iface = "buildcraft.api.tiles.IHasWork", modid = Constants.BCLIB_MOD_ID)
+public abstract class TilePowered extends TileBase implements IRenderableTile, ISpeedUpgradable, IStreamableGui, IHasWork {
```
AlexIIL: Just FYI `IHasWork` is now obtained through the capability, rather than the interface. I think the javadoc might be out of date if it doesn't specify that now.

temp1011:
If you check on server side then you can use `getInternalInventory()` on `TileMailBox` which might have something useful in. Might be difficult to look through though.

src/main/java/forestry/core/tiles/TilePowered.java
dblsaiko: Ah, okay. No, the javadoc doesn't specify that. Thanks for the heads up!

dblsaiko:
@temp1011 I've never used the mail system, how do I check if this works?

temp1011:
I'm not sure, I've never really used mail. The inventory can contain things like blank letters though so I think you have to check specific slots. Which ones I have no idea. I think there should be a better way than this

dblsaiko:
Hmm, there's POBoxInfo#hasMail, maybe that's the one I want.
EDIT: Hmm, I'd need a player to get an instance of that though, I think. :/

temp1011:
You might be able to get to that though `getContainer()`. But, in fact, I didn't read the code carefully enough. It seems that the internal inventory might be for letters only after all? Maybe I'm still not reading things right though.
https://github.com/ForestryMC/ForestryMC/blob/7df3a949165fa1d15651228c1a1c65b6ebfb0769/src/main/java/forestry/mail/tiles/TileMailbox.java#L38

dblsaiko:
Yeah I can get it through getContainer(), that needs a player though. Hmm

EDIT: I just sent a letter to myself, and while it did show up in the mailbox, the internal inventory was still empty, so this isn't what I want.

build.gradle
```diff
@@ -108,8 +111,9 @@ jar {
 }
 
 task sourcesJar(type: Jar) {
-	from zipTree('build/retromapping/retromappedReplacedMain.jar') // Not sure if this is how Abrar designed it, but grab the remapped src
+	// Not sure if this is how Abrar designed it, but grab the remapped src
```
mezz: Why change this? The comment is on the wrong line now...

build.gradle
```diff
@@ -63,6 +65,7 @@ dependencies {
 	runtime "mezz.jei:jei_${jei_mcversion}:${jei_version}"
 	deobfCompile "net.industrial-craft:industrialcraft-2:${ic2_version}:api"
 	deobfCompile "TechReborn:TechReborn-${TR_mcversion}:${TR_version}:api"
+	deobfCompile "com.mod-buildcraft:buildcraft:${buildcraft_version}"
```
mezz: Is there an API we can compile against? I'd like to avoid compiling against their internal classes by accident.

src/main/java/forestry/apiculture/ModuleApiculture.java
```diff
@@ -944,7 +943,7 @@ public ISaveEventHandler getSaveEventHandler() {
 
 	@Override
 	public void populateChunk(IChunkGenerator chunkGenerator, World world, Random rand, int chunkX, int chunkZ,
-			boolean hasVillageGenerated) {
+														boolean hasVillageGenerated) {
```
mezz: ???

src/main/java/forestry/core/config/EnergyDisplayMode.java
```diff
@@ -0,0 +1,24 @@
...
+	private final String energyName;
+	private final float factor;
+
+	EnergyDisplayMode(String energyName, float factor) {
+		this.energyName = energyName;
+		this.factor = factor;
+	}
+
+	public String formatRate(int energy) {
+		return String.format("%.0f %s/t", energy * factor, energyName);
```
mezz: This class should use a localized number format so that it matches the language set in the game instead of the language of the system.
```java
Locale locale = MinecraftForgeClient.getLocale();
NumberFormat numberFormat = NumberFormat.getIntegerInstance(locale);
float amount = energy * factor;
String amountString = numberFormat.format(amount);
return String.format("%s %s/t", amountString, energyName);
```

src/main/java/forestry/core/tiles/TileBase.java
```diff
@@ -41,4 +44,10 @@ public boolean shouldRefresh(World world, BlockPos pos, IBlockState oldState, IB
 		Block newBlock = newState.getBlock();
 		return oldBlock != newBlock || !(oldBlock instanceof BlockBase) || !(newBlock instanceof BlockBase);
 	}
+
+	@Nonnull
+	public EnumFacing getFacing() {
+		return EnumFacing.getFront(getBlockMetadata());
```
mezz: I think this can be done more properly using the blockstate properties. Block metadata is going away in 1.13 so we should avoid using it if possible.

src/main/java/forestry/core/tiles/TileEngine.java
```diff
@@ -303,15 +300,18 @@ public EnergyManager getEnergyManager() {
 
 	@Override
 	public boolean hasCapability(Capability<?> capability, @Nullable EnumFacing facing) {
-		return energyManager.hasCapability(capability) || super.hasCapability(capability, facing);
+		return (getFacing() == facing && energyManager.hasCapability(capability)) ||
```
mezz: Engines can receive energy so that they can chain, I think this change will break it.

src/main/java/forestry/core/triggers/Trigger.java
```diff
@@ -10,62 +10,84 @@
...
+	private final String textureName;
+
+	protected Trigger(String uid) {
+		this(uid, uid);
+	}
+
+	protected Trigger(String uid, String localization) {
+		this.uid = "forestry:" + uid;
+		this.localization = localization;
+		this.textureName = localization.replaceAll("([a-z])([A-Z])", "$1_$2").toLowerCase();
```
mezz: String transformations like this make things harder to find.
Can you pass in the texture name in to the constructor as a normal string instead?

src/main/java/forestry/core/triggers/Trigger.java
```diff
@@ -10,62 +10,84 @@
...
+import net.minecraftforge.fml.relauncher.SideOnly;
+
+import javax.annotation.Nullable;
+
+import buildcraft.api.core.render.ISprite;
+import buildcraft.api.statements.IStatement;
+import buildcraft.api.statements.IStatementParameter;
+import buildcraft.api.statements.ITriggerExternal;
+import buildcraft.api.statements.StatementManager;
+import buildcraft.lib.client.sprite.SpriteRaw;
```
mezz: We should avoid importing things from buildcraft that are not in the API. I'd rather reimplement our own `SpriteRaw` if necessary.

src/main/java/forestry/energy/EnergyManager.java
```diff
@@ -1,29 +1,29 @@
...
-import net.minecraftforge.energy.IEnergyStorage;
-
+import buildcraft.api.mj.*;
 import forestry.api.core.INbtReadable;
 import forestry.api.core.INbtWritable;
+import forestry.core.config.Config;
 import forestry.core.network.IStreamable;
 import forestry.core.network.PacketBufferForestry;
 import forestry.energy.compat.EnergyStorageWrapper;
+import forestry.energy.compat.mj.*;
```
mezz: Please avoid `*` imports

src/main/java/forestry/energy/compat/mj/MjConnectorWrapper.java
```diff
@@ -0,0 +1,22 @@
...
+public class MjConnectorWrapper implements IMjConnector {
+	protected final EnergyManager energyManager;
+
+	public MjConnectorWrapper(EnergyManager energyManager) {
+		this.energyManager = energyManager;
+	}
+
+	@Override
+	public boolean canConnect(@Nonnull IMjConnector other) {
+		return true; // TODO
```
mezz: This should be fixed before merging

src/main/java/forestry/plugins/PluginBuildCraftFuels.java
```diff
@@ -10,41 +10,57 @@
...
 import forestry.modules.ForestryModuleUids;
 
-// TODO: Buildcraft for 1.9
 @ForestryModule(containerID = ForestryCompatPlugins.ID, moduleID = ForestryModuleUids.BUILDCRAFT_FUELS, name = "BuildCraft 6 Fuels", author = "mezz", url = Constants.URL, unlocalizedDescription = "for.module.buildcraft6.description")
 public class PluginBuildCraftFuels extends BlankForestryModule {
 
 	@Override
 	public boolean isAvailable() {
-		return ModUtil.isAPILoaded("buildcraft.api.fuels", "[2.0, 3.0)");
+		return ModUtil.isModLoaded(Constants.BCLIB_MOD_ID);
```
mezz: We should give this a version range so that it is inactive on the next major version.

src/main/java/forestry/plugins/PluginBuildCraftFuels.java
```diff
@@ -10,41 +10,57 @@
...
+		if (ethanol != null) {
+			long ethanolPower = 40 * MjAPI.MJ;
+			int ethanolBurnTime = Math.round(Constants.ENGINE_CYCLE_DURATION_ETHANOL * ForestryAPI.activeMode.getFloatSetting("fuel.ethanol.combustion"));
+			BuildcraftFuelRegistry.fuel.addFuel(ethanol, ethanolPower, ethanolBurnTime);
+		}
+	}
+
+	@Override
+	public void doInit() {
+		// FIXME postInit belongs in here, really
```
mezz: FIXME

src/main/java/forestry/plugins/PluginBuildCraftRecipes.java
```diff
@@ -10,33 +10,33 @@
...
-
+import net.minecraftforge.fml.common.Optional;
 
 @ForestryModule(containerID = ForestryCompatPlugins.ID, moduleID = ForestryModuleUids.BUILDCRAFT_RECIPES, name = "BuildCraft 6 Recipes", author = "SirSengir", url = Constants.URL, unlocalizedDescription = "for.module.buildcraft6.description")
 public class PluginBuildCraftRecipes extends BlankForestryModule {
 
 	@Override
 	public boolean isAvailable() {
-		return ModUtil.isAPILoaded("buildcraft.api.recipes", "[2.0, 4.0)");
+		return ModUtil.isModLoaded(Constants.BCLIB_MOD_ID);
```
mezz: needs version range

src/main/java/forestry/plugins/PluginBuildCraftRecipes.java
```diff
@@ -10,33 +10,33 @@
...
 
-	@Optional.Method(modid = "BuildCraftAPI|recipes")
+	@Optional.Method(modid = Constants.BCLIB_MOD_ID)
 	@Override
 	public void registerRecipes() {
 		// Add recipe for ethanol
-		// TODO: Buildcraft for 1.9
-//		BuildcraftRecipeRegistry.refinery.addRecipe("forestry:BiomassToEthanol", Fluids.BIOMASS.getFluid(4), Fluids.BIO_ETHANOL.getFluid(1), 100, 1);
+		BuildcraftRecipeRegistry.refineryRecipes.addDistillationRecipe(Fluids.BIOMASS.getFluid(4),
+				Fluids.BIO_ETHANOL.getFluid(0), Fluids.BIO_ETHANOL.getFluid(1), 100);
```
mezz: Why 0 fluid?

src/main/java/forestry/plugins/PluginBuildCraftStatements.java
```diff
@@ -10,57 +10,72 @@
...
 @ForestryModule(containerID = ForestryCompatPlugins.ID, moduleID = ForestryModuleUids.BUILDCRAFT_STATEMENTS, name = "BuildCraft 6 Statements", author = "mezz", url = Constants.URL, unlocalizedDescription = "for.module.buildcraft6.description")
-//@Optional.Interface(iface = "buildcraft.api.statements.ITriggerProvider", modid = "BuildCraftAPI|statements")
-public class PluginBuildCraftStatements extends BlankForestryModule {//implements ITriggerProvider {
+@Optional.Interface(iface = "buildcraft.api.statements.ITriggerProvider", modid = Constants.BCLIB_MOD_ID)
+public class PluginBuildCraftStatements extends BlankForestryModule implements ITriggerProvider {
 
 	@Override
 	public boolean isAvailable() {
-		return ModUtil.isAPILoaded("buildcraft.api.statements", "[1.0, 2.0)");
+		return ModUtil.isModLoaded(Constants.BCLIB_MOD_ID);
```
mezz: version range

src/main/java/forestry/energy/EnergyManager.java
```diff
@@ -1,29 +1,29 @@
...
-
-import net.minecraft.nbt.NBTTagCompound;
-import net.minecraft.util.math.MathHelper;
-
-import net.minecraftforge.common.capabilities.Capability;
-import net.minecraftforge.energy.CapabilityEnergy;
-import net.minecraftforge.energy.EnergyStorage;
-import net.minecraftforge.energy.IEnergyStorage;
-
+import buildcraft.api.mj.*;
```
mezz: Please avoid `*` imports

src/main/java/forestry/energy/compat/mj/MjHelper.java
```diff
@@ -0,0 +1,104 @@
+package forestry.energy.compat.mj;
+
+import buildcraft.api.mj.*;
```
mezz: Please avoid `*` imports

mezz (changes requested):
Looking good, I think we're close.

src/main/java/forestry/plugins/PluginBuildCraftRecipes.java
dblsaiko: It crashes with a null fluid, that parameter is the output for the gas (the top). Talked about this with @AlexIIL in the Forestry discord, in case you want to read it

build.gradle
dblsaiko: Thought I fixed this... Whoops

build.gradle
dblsaiko: BuildCraft Lib should work, right? There's no seperate API package

src/main/java/forestry/core/tiles/TileBase.java
dblsaiko: Used that because it caches the result, but I agree it's better to use the blockstate property.

src/main/java/forestry/energy/compat/mj/MjConnectorWrapper.java
dblsaiko: BC classes return true here too, I'm just going to remove the TODO :P

dblsaiko:
What should I do with TriggerHasMail? I assume it was meant for the mailbox at some point, but that stores its inventory per player

dblsaiko:
Everything you said should be fixed now.

mezz (changes requested):
You can just leave out `TriggerHasMail` for now.  I think the mail box is implemented in a weird way because of how I ported it to 1.7.10. It should be one person's mailbox, and anyone else who tries to open it should be locked out. The way it works now is whoever tries to open the mailbox will see their own mailbox's inventory. There is no way for the trigger to work. Fixing it should come after this PR in a separate commit.

One thing I thought of that you should address is MJ to RF conversion. Right now an engine can accept either energy type, and output either energy type. This may be against the wishes of Buildcraft. Please check with them to see if the conversion is meant to be that simple. If they do not think it should be automatically convertible to RF (this is probably the case, it seems bad), we should use a config switch that either makes the engines MJ-only-mode or RF/Tesla/FE mode.

dblsaiko:
Yup, so energy conversion is bad. I'll add a config option for that

dblsaiko:
Actually, wouldn't it be a good idea to only allow power input to engines from other engines to still allow chaining, but don't allow stuff like power pipes connecting to it? That would fix the problem and engines would still be usable for both MJ and RF generation.

mezz:
That sounds like a good solution to me.

dblsaiko:
Engine chaining looks like it's special cased, actually, I don't think I need to write any special handling code... In fact, it should already not accept energy from pipes right now

dblsaiko:
No idea why I didn't test engine chaining last time before ripping it out again

dblsaiko:
I just saw that the Still converts 10 mB of Biomass to 3 mB of Ethanol, should I adjust the values the BC Distiller gives to that?

mezz (approved):
Looks good to me

PR #2181, opened 2018-07-07
fix cocoon tile name

temp1011:
This follows the same pattern as https://github.com/ForestryMC/ForestryMC/commit/9b4a4c2b1d0b495f180a4dfea2bcfe492546b4d0 and just finishes cleaning up tile registration names. If someone could let me know how to get the cocoon to test this it would be appreciated :smile: .


temp1011:
Looks like this change is the same one that the candle underwent in the previous commit

Nedelosk:
@temp1011 Can you please add it to `MigrationHelper`?

temp1011:
Hopefully this is right

Nedelosk:
Looks right, thanks

PR #2197, opened 2018-07-17
add methods to allow formatting config comments

temp1011:
This came up while working on something else. Adds methods to `LocalizedConfiguration` to allow formatted comment keys. Also removes canTranslate check. I think it's better if something formatted is displayed rather than just the key. However, I can revert this.

Common config after this: https://gist.github.com/temp1011/6dc8d96304a7647f40eaa27f3dd85304.


mezz:
I see some errors with reporting the defaults when looking at a diff of the old and new configs. Please take a close look that everything matches except for the config comments.

mezz (changes requested):
I see some differences in the diff, like missing "default" or incorrect arrays in the defaults. Please make sure the only changes are to the config comments.

temp1011:
[new config](https://gist.github.com/temp1011/2ff1bfad740000ac8d41fb4d9acdfd24). I think this now has no differences with the old config apart from the addition of some comments.

src/main/java/forestry/core/config/LocalizedConfiguration.java
```diff
@@ -115,43 +111,55 @@ public String getStringLocalized(String category, String name, String defaultVal
...
 
 		Property prop = this.get(category, name, defaultValue);
 		prop.setLanguageKey(langKey);
 		prop.setValidValues(validValues);
-		prop.setComment(comment + " [default: " + Arrays.toString(defaultValue) + "] [valid: " + Arrays.toString(prop.getValidValues()) + "]");
+		prop.setComment(comment + getValiedOptions(prop));
 		return prop.getStringList();
 	}
 
+	private String getValiedOptions(Property prop) {
```
mezz: method name has a typo

mezz (changes requested):
looks good except for a small typo

src/main/java/forestry/core/config/LocalizedConfiguration.java
temp1011: oops, sorry

mezz (approved):

PR #2204, opened 2018-07-23
allow hive worldgen to scale 

temp1011:
at the moment the numbers in `HiveDecorator` are fairly hardcoded. While working on https://github.com/ForestryMC/Binnie/pull/486 I realised that this makes it hard to get decent numbers of every hive. If this will make the world too busy with hives we can just ignore it but there won't be as many Binnies hives. Let me know what you think.


mezz (approved):
It makes sense that you should get more hives in your world if you add more Forestry addons with hives.
That is how it worked when each mod created hives independently.
Looks good to me!

PR #2216, opened 2018-08-06
remove duplicate call to read IMC messages

temp1011:
Probably best to read the IMC messages in the event. I believe the second call would have returned an empty list anyway.


temp1011:
hmm perhaps I'm wrong here. From what I can tell both calls do the same thing though. Maybe not all messages are sent at the first point.

PR #2217, opened 2018-08-09
fix wrong drops from orchard with harvestcraft closes #2211

### LINKED ISSUE #2211: Orchard not harvesting Pam's fruit correctly
Repost of a comment in #1718 as it's probably a different bug, so it gets it's own open issue:

Something with Pam's fruits seems to be broken again, the orchard mode farm is harvesting the actual fruit 'nodes', instead of collecting the products. It oddly enough replaces the harvested nodes too.

https://user-images.githubusercontent.com/2611674/43587483-ace44b0e-96ad-11e8-8bee-893afe8ce7a6.png

The screenshot shows the harvested nodes in the farm output, instead of the fruit.

Tested in FTB Continuum 1.4.1, with forestry updated to forestry_1.12.2-5.8.1.326

temp1011:
Harvestcraft overrides the deprecated version of `getDrops()`. So the crop must too. The deprecated version normally points to the normal version anyway so there shouldn't be any regressions.

### DIFF
--- a/src/main/java/forestry/farming/logic/crops/CropDestroy.java
+++ b/src/main/java/forestry/farming/logic/crops/CropDestroy.java
@@ -12,6 +12,7 @@ package forestry.farming.logic.crops;
 
 import javax.annotation.Nullable;
 import java.util.Iterator;
+import java.util.List;
 
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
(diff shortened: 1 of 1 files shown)


temp1011:
PR to fix in harvestcraft https://github.com/MatrexsVigil/harvestcraft/pull/255.

Issue #1973, opened 2018-02-05
[Feature Request] Please add the option to disable certain blocks completely


KalaMiDeviL:
I'm using this mod alongside other big tech mods (as most people do) and currently have 3 kinds of stills, 3 machines for filling and another 3 for emptying containers. I'd like to remove all but the Thermal Expansion machines - especially the 1000-recipe-recipe trees they each dump into JEI. Could you make that an option?

PR #2221, opened 2018-08-12
 add config for disabling factory machines 

temp1011:
Closes https://github.com/ForestryMC/ForestryMC/issues/1973.

The changes seem very heavy but I'm not sure of a better way to add this in. Hopefully in 1.13 this could be rewritten to reduce the code needed.

There are 2 config options for each machine (aside from the bottler) since it's possible someone might use the forestry recipe map to create recipes for their machine. However, it's probably a bit overkill so I could remove it.

### DIFF
--- a/src/main/java/forestry/api/fuels/DummyMap.java
+++ b/src/main/java/forestry/api/fuels/DummyMap.java
@@ -1,0 +1,4 @@
+package forestry.api.fuels;
+
+public class DummyMap {
+}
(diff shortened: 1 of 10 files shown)


Nedelosk:
I think it would make more sense and would be much cleaner to create one module for every machine and keep the "factory" module as a dependency. What do you think @mezz ?

mezz:
Lex and cpw always recommend registering blocks even if they are disabled. That makes config differences more compatible between clients and servers, and it avoids block ID churn from enabling and disabling things while a world is active.

The right thing to do is just disable the recipes, JEI integration, etc, so that the machines are not obtainable in-game.

Splitting "factory" into tiny mods or modules is pretty good option, but I think it would make things a bit too complicated.

temp1011:
The blocks will now always be registered but won't be visible in JEI or have any recipes registered.

mezz (approved):

PR #2231, opened 2018-08-20
Check for IToolGrafter instead of ItemGrafter for non-Forestry leaves

dmarcuse:
There's currently a minor issue where items implementing `IToolGrafter` work for Forestry leaves, but not for vanilla or other mod leaves. This is because [AbstractLeaves checks for IToolGrafter](https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/src/main/java/forestry/arboriculture/blocks/BlockAbstractLeaves.java#L179), while [ModuleArboriculture.onHarvestDropsEvent checks for ItemGrafter](https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/src/main/java/forestry/arboriculture/ModuleArboriculture.java#L615).


dmarcuse:
Thank you!

Issue #2238, opened 2018-08-31
[Suggestion] Charcoal pit config


Agrax:
I would like to be able to nerf charcoal pit a tad bit. Dirt as a fireproof material giving 12 charcoal per wood pile is kinda OP. Ik i can very easily just tweak how much wood does a wood pile need via CT, but imho we need a tiny bit better scaling in our modpack. Dirt<netherrack. Could you please consider adding such a config, please?

temp1011:
See https://github.com/jaredlll08/ModTweaker/pull/678 for the modtweaker side of things.

Agrax:
Works smoothly. Thanks

PR #2240, opened 2018-09-04
add methods to allow removing charcoal walls

temp1011:
See #2238. This could be a config but anyone else can register their own walls which makes it difficult to handle. I think it's better instead to add methods to remove walls (even if they are not guaranteed) and then expose this through modtweaker (I will write this for all the charcoal manager methods).


Rongmario:
As well as being able to remove existing walls, registerWall should be exposed in ModTweaker too. Also, is ash output dependent on charcoal output?

temp1011:
I am planning to add modtweaker methods for `registerWall` too. From the drop code it looks like ash and charcoal output are linked: https://github.com/ForestryMC/ForestryMC/blob/6aec5b6f25b9cf2eb9435cb8a012a8a16b4a1cc1/src/main/java/forestry/arboriculture/blocks/BlockAsh.java#L71-L78

Rongmario:
Ah, thought so; makes sense now. I like that design choice but for others it may be better to make ash output customizable.

PR #2248, opened 2018-09-09
more fixes for charcoal

temp1011:
- use null instead of air blocks in CharcoalPileWall to prevent matching returning true for air blocks:
Caused a regression where charcoal pits with invalid walls would just match the first wall because air state would match with any walls registered with just a block.

- don't give any charcoal for blocks that aren't charcoal walls:
The old code (#2241 and before) gave at least 1 charcoal per face but I don't think this is balanced, since charcoal piles then just become a furnace for free that give ash too. Also updates `BlockAsh` to use the new `getDrops` method.

### DIFF
--- a/src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
+++ b/src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
@@ -201,7 +201,11 @@ public class BlockWoodPile extends Block implements IItemModelRegister, IStateMa
 		for (EnumFacing facing : EnumFacing.VALUES) {
 			charcoalAmount += getCharcoalFaceAmount(world, pos, facing);
 		}
-		return charcoalAmount / 6;
+		float ret = charcoalAmount / 6;
+		if (ret > 0F) {
+			ret += 9F;    //base charcoal drop amount for ash block, calculated here instead so invalid walls son't still drop charcoal
+		}
+		return Math.min(ret, 15.0F);
 	}
 
 	private int getCharcoalFaceAmount(World world, BlockPos pos, EnumFacing facing) {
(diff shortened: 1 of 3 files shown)


src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
```diff
@@ -201,7 +201,11 @@ private float getCharcoalAmount(World world, BlockPos pos) {
 		for (EnumFacing facing : EnumFacing.VALUES) {
 			charcoalAmount += getCharcoalFaceAmount(world, pos, facing);
 		}
-		return charcoalAmount / 6;
+		float ret = charcoalAmount / 6;
+		if (ret > 0F) {
+			ret += 9F;    //base charcoal drop amount for ash block, calculated here instead so invalid walls son't still drop charcoal
```
Nedelosk: What if the ret is at this moment bigger than 6? The ash block only has a valid metadata from 0 to 15.

Nedelosk (changes requested):

src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
temp1011: I'm not sure what would happen but it probably wouldn't be good. Hopefully fixed.

src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
```diff
@@ -201,7 +201,11 @@ private float getCharcoalAmount(World world, BlockPos pos) {
 		for (EnumFacing facing : EnumFacing.VALUES) {
 			charcoalAmount += getCharcoalFaceAmount(world, pos, facing);
 		}
-		return charcoalAmount / 6;
+		float ret = charcoalAmount / 6;
+		if (ret > 0F) {
+			ret += 9F;    //base charcoal drop amount for ash block, calculated here instead so invalid walls won't still drop charcoal
```
Nedelosk: Why is this at `BlockWoodPile.getCharcoalAmount` and not in `BlockAsh.getDrops` after `if (amount > 0)` ?

Nedelosk (changes requested):

src/main/java/forestry/arboriculture/blocks/BlockWoodPile.java
temp1011: Not sure, probably I got confused. Having it in this method changes behaviour too because the amount gets clamped at the end.

Nedelosk (approved):

PR #2249, opened 2018-09-12
update API comments regarding use of null

temp1011:
The API comment says that null should be allowed for binomial: https://github.com/ForestryMC/ForestryMC/blob/675211df52c4a69e9c2e1f39d68cba8b133b999d/src/main/java/forestry/api/apiculture/IBeeFactory.java#L24


Nedelosk:
I don't think it would be good to have `null` as a valid value. It would just confuse people. Please just remove the `or null` part of the comment. And if you do so please remove the `(may be null)` and `or return null` part from the `getBinomial()` and `getAuthority()` comments.

temp1011:
OK, I have updated the comments. I agree this makes more sense.

PR #2253, opened 2018-09-18
clean up init of allelehelper

temp1011:
IMO it's better to keep everything inside the class


MatthiasMann:
Won't this run the init code any time getInstance() is called?

temp1011:
Yep, somehow I messed this up, its on the wrong line...

PR #2265, opened 2018-10-07
Refine some models for filesize and performance

Maluscat:
I encountered some models which could use some improvements and improved them:
- Improved the wood pile with cullfacing and disabling not visible faces
- Made a parent for the one-block-farms so no code is copy-pasted for every farm
- Improved the database model with cullfacing

I also have a question: how are the blocks like the bottler or squeezer modeled? One could also improve them a bit

### DIFF
--- a/src/main/resources/assets/forestry/models/block/arboretum.json
+++ b/src/main/resources/assets/forestry/models/block/arboretum.json
@@ -1,57 +1,7 @@
 {
-	"parent": "block/block",
+	"parent": "forestry:block/farm_parent",
 	"textures": {
 		"particle": "forestry:blocks/arboretum_particle",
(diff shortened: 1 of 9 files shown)


temp1011:
I don't really have any experience with models but it looks like the factory machines are specified as alveary variants? [link](https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/src/main/resources/assets/forestry/blockstates/alveary.json). But that seems to give links to other stuff so I'm probably wrong.

Nedelosk:
The models of the factory machines are generated in code. Seems like I never cleaned up the [alveary.json](https://github.com/ForestryMC/ForestryMC/blob/mc-1.12/src/main/resources/assets/forestry/blockstates/alveary.json) file.

Issue #2264, opened 2018-10-07
Charcoal Pit causes increasing memory usage until the game freezes/crashes


gwpantazes:
See [the log here](https://gist.github.com/gwpantazes/7e7cab12c7d4281e4fe1a6912d0422d1).

Reproducible in this environment:

- Minecraft 1.12.2 (modded) forge-14.23.4.2760
- Stoneblock by kasparsu Version 1.0.17
   - Forestry - forestry+1.12.2-5.8.1.344.jar
   - 5493 MB memory allocated

> Try the latest version, it may have fixed the issue already.

Sorry, I'm not sure how to try the most recent version, but thought I'd file this issue anyway just so that this is a known issue. I tried searching existing issues, and this could possibly be related to #1878 because it has to deal with Charcoal Pit, but since this issue causes a crash I consider this more critical of a bug.

My initial guess of what's happening from a black-box perspective is that the charcoal pit functionality has a memory leak or something. The game runs fine as long as I don't use the charcoal pit functionality.

# Steps to Reproduce
Note: I'm framing this from the Stoneblock modpack perspective because that's where it's reproducing the bug for me. If this bug reproduces in other packs or from a simply vanilla+forestry environment, can someone please update here?

1. Boot up a new Stoneblock world with cheats/creative
2. Make a 2 width, 3 long, 1 tall charcoal pit out of clay
3. Put in the woodpiles, light the corner woodpile with a flint and steel, and seal up the charcoal pit
4. Let it burn / wait.
5. Observe that after a while, the game will start to stagger / slow, and eventually will freeze and crash. See [the log](https://gist.github.com/gwpantazes/7e7cab12c7d4281e4fe1a6912d0422d1) for the error messages.

gwpantazes:
Attaching a timelapse video of the memory usage from around the time I seal up the charcoal pit until the crash. It seems like the memory issue only becomes a problem when the charcoal pit is becoming ash.

[Charcoal Pit Memory Crash Timelapse.mp4.zip](https://github.com/ForestryMC/ForestryMC/files/2453572/Charcoal.Pit.Memory.Crash.Timelapse.zip)

gwpantazes:
Also tried in FTB Revelation to compare, and couldn't reproduce. Even with less RAM allocated overall, the game didn't crash when using the charcoal pit.

FTB Revelation Version 2.5.0
Minecraft Version: 1.12.2 (modded) forge-14.23.4.2759
[...]

PR #2268, opened 2018-10-09
prevent wood pile from loading more chunks

temp1011:
I think this is the cause of https://github.com/ForestryMC/ForestryMC/issues/2264. In the stoneblock map there aren't any air blocks so a charcoal pit without a wall built next to the edge could continue scanning chunks forever. I thought this could occur when in https://github.com/ForestryMC/ForestryMC/pull/2241 but I couldn't get it to happen there.

Another solution would be to limit the maximum size of the charcoal pit.


Nedelosk:
This is probably not the cause of #2264 because that would cause a `StackOverflowError` and not an `OutOfMemoryError`.

temp1011:
In the latest crash report on the issue there is a lot about generating a new chunk though, which would suggest that the pile is loading new chunks.

PR #2272, opened 2018-10-14
prevent overflow of slots in database causing contents to be lost

temp1011:
Thanks @mconwa01 for finding the bug.

While doing this I also noticed there is a bug where the last slot of the database won't render the item in it. I tried changing this code down by 1 and the item would render but the slot would be in the wrong place: https://github.com/ForestryMC/ForestryMC/blob/7a45b02ca64b3bb952c2723e4f9868f3fcdda80d/src/main/java/forestry/database/gui/GuiDatabase.java#L147-L148

Presumably there is an off by 1 error somewhere for the rendering bug but I can't find it yet.

### DIFF
--- a/src/main/java/forestry/core/utils/InventoryUtil.java
+++ b/src/main/java/forestry/core/utils/InventoryUtil.java
@@ -617,7 +617,8 @@ public abstract class InventoryUtil {
 
 	/**
 	 * The database has an inventory large enough that int must be used here instead of byte
-	 */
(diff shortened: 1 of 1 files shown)


temp1011:
Yep you're right, this does break everything. I guess the options are to limit the database to 127 slots or add some kind of migration.

temp1011:
Hopefully should be fixed now

PR #2288, opened 2018-11-04
Fix #239 Change Yield behavior from drop chance to leafe chance

### LINKED ISSUE #239: Ye Olde Orchard Bug
Forestry Build 190 on Forge 1217 / Minecraft 1.7.10, no other mods installed. Created this instance just to test for this issue.

This is a bug that has existed throughout 1.6.x and never got fixed due to Forestry development being dormant for a period of time. Apparently it still lives on in 1.7.x.

Issue: the orchard mode manual multifarm sometimes does not generate output when harvesting. Oddly enough, this seems tree-specific. For example, papaya orchards have never proven to be a problem for me even once, but on the other hand, any lemon orchard I've tried barely has any output. People have also reported this for walnut (or maybe chestnut?) trees, but I've been unable to verify this consistently; maybe it's even trait-related, and their walnut trees had different traits than mine.

Reproduction: Build multifarm. Program for orchard with copper tubes in manual mode. Lay out farmland. Plant and grow lemon trees. Bonemeal leaves to instantly ripen the fruit. Observe the farm harvesting the ripe fruit. Check farm inventory for how many lemon items are actually created compared to how many ripe fruit blocks were harvested.

In my test just now, I bonemealed 17 leaves, all of which were dutifully harvested, and I got 3 lemons out of that. If I were to try this with papayas, I would receive 17 fruit for 17 harvest actions, as expected.

Nedelosk:
This fixes an old issue that was created 4 years ago.
Yield now define how many fruit leaves/pods a tree contains all other
leaves are leaves without any fruits. All ripe leaves now have a 100 %
chance of a fruit drop and can be right clicked to harvest the fruit.


mezz:
This looks very nice!
I will test it out to see.

mezz (changes requested):
Looks great.
Only issue I found in testing is that worldgen trees (I found a cherry tree) still generate with 100% fruit leaves.

Nedelosk:
@mezz Fixed.

mezz:
Nice work @Nedelosk !

Issue #2323, opened 2018-12-22
Ticking World Exception


joker-119:
// On the bright side, I bought you a teddy bear!

Time: 12/22/18 3:17 AM
Description: Exception ticking world

java.lang.NullPointerException: Exception ticking world
	at net.minecraft.world.chunk.Chunk.func_177436_a(Chunk.java:542)
	at net.minecraft.world.World.func_180501_a(World.java:343)
	at net.minecraft.world.World.func_175656_a(World.java:425)
	at forestry.arboriculture.tiles.TileLeaves.onBlockTick(TileLeaves.java:154)
	at forestry.arboriculture.blocks.BlockForestryLeaves.func_180650_b(BlockForestryLeaves.java:100)
	at net.minecraft.block.Block.func_180645_a(Block.java:508)
	at net.minecraft.world.WorldServer.func_147456_g(WorldServer.java:476)
	at net.minecraft.world.WorldServer.func_72835_b(WorldServer.java:225)
	at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:756)
	at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:397)
	at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:668)
	at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:526)
	at java.lang.Thread.run(Thread.java:748)

It seems to happen pretty randomly. Sometimes just while running around, sometimes while loading new chunks, sometimes when teleporting. Doing the same things doesn't always cause the crash again (IE: Teleporting from one player to another can cause the crash, but once we login, if I go back then teleport to them again, nothing.

Nedelosk:
Please update Binnies mods.

Nedelosk:
Duplicate of #2317

PR #2326, opened 2018-12-23
prevent crash for leaves where default can't be found

temp1011:
Stops the crash in #2323 but doesn't necessarily fix the underlying issue. Not sure what causes the crash so I've just added a logging diagnostic to hopefully work it out.

### DIFF
--- a/src/main/java/forestry/Forestry.java
+++ b/src/main/java/forestry/Forestry.java
@@ -79,7 +79,8 @@ import forestry.plugins.PluginTechReborn;
 		+ "after:" + PluginNatura.MOD_ID + ";"
 		+ "after:toughasnails;"
 		+ "after:" + PluginTechReborn.MOD_ID + ";"
-		+ "after:" + PluginBuildCraftFuels.MOD_ID + ";")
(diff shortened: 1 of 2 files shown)


temp1011:
Is it worth leaving this open if it's a bug in Binnie's mods? Still might be nice to have some safety against other mods causing crashes?

Nedelosk:
If the people update forestry but not binnie they will get this crash because forestry can't find the fruit leaves that were added with the last binnie update. Would be great if you could change it to :
```
IBlockState defaultLeaves = null;
if (isFruitLeaf) {
   defaultLeaves = ModuleArboriculture.getBlocks().getDefaultLeavesFruit(primary.getUID());
} 
if(defaultLeaves == null) {
   defaultLeaves = ModuleArboriculture.getBlocks().getDefaultLeaves(primary.getUID());
}
worldIn.setBlockState(getPos(), defaultLeaves);
```
Additionaly it would be great if you could add `"before:binniecore@[2.5.1.184,)"` to the dependencies of forestry in `Forestry.java`.

PR #2327, opened 2018-12-23
fix #2324 habitat former can't lower temperature

### LINKED ISSUE #2324: habitat former cannot lower temperature
Habitat former cannot lower temperature below minecraft ambient for whatever biome it is placed in.  It can raise the temperature correctly to any level above ambient, and acknowledges whatever setting change made, but just absolutely cannot lower it.  Machine is fully powered, and tried all three acceptable liquids (despite instructions stating power is for temp adjustment, liquids are for humidity), but no lowering effect can be achieved.  Stripped minecraft down to just the most recent version of the Forestry mod.  Tested in normal minecraft world "plains" biome - ambient 79temp, 40humidity, and superflat world ambient 80temp, 40humidity.  Humidity control is fully 0-100% functional, temperature can be elevated to 200, and then lowered again to noraml, but never lowered below biome ambient.  This issue makes it impossible to breed/raise cold climate bees without physically setting up "camp" in the actual biome the queens require.  Was really hoping to breed some "Merry Bees" before their Dec27 time window closes.

Forge 2768
Forestry 1.12.2-5.8.2.357
Win10-64
Java 1.8.0_191

temp1011:
The code looks a bit ugly but does work.

### DIFF
--- a/src/main/java/forestry/core/climate/ClimateTransformer.java
+++ b/src/main/java/forestry/core/climate/ClimateTransformer.java
@@ -117,6 +117,7 @@ public class ClimateTransformer implements IClimateTransformer, IStreamable, INb
 	@Override
 	public IClimateManipulatorBuilder createManipulator(ClimateType type) {
 		return new ClimateManipulator.Builder()
+					.setAllowBackwards()
 			.setDefault(defaultState)
 			.setCurrent(currentState)
 			.setTarget(targetedState)
--- a/src/main/java/forestry/core/climate/ClimateTransformer.java
+++ b/src/main/java/forestry/core/climate/ClimateTransformer.java
@@ -117,7 +117,7 @@ public class ClimateTransformer implements IClimateTransformer, IStreamable, INb
 	@Override
 	public IClimateManipulatorBuilder createManipulator(ClimateType type) {
 		return new ClimateManipulator.Builder()
-					.setAllowBackwards()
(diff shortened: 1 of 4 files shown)


Nedelosk:
This would only break the system even more. The only change you need to make is in `ClimateManipulator.Builder#setAllowBackwards`. You only have to set `backwards` to`true` instead of `false`.

src/main/java/forestry/core/climate/ClimateTransformer.java
```diff
@@ -117,6 +117,7 @@ public void readFromNBT(NBTTagCompound nbt) {
 	@Override
 	public IClimateManipulatorBuilder createManipulator(ClimateType type) {
 		return new ClimateManipulator.Builder()
+				.setAllowBackwards()
```
Nedelosk: This breaks the humidity implementation of the habitat former because the humidity should only be controlled through the fluids and this allows to lower the humidity by consuming water which makes no sense. There already is a `setAllowBackwards()` in `TileHabitatFormer#updateTemperature` for the regulation of the temperature.

Nedelosk (changes requested):

src/main/java/forestry/core/climate/ClimateTransformer.java
temp1011: Yep, my mistake.

PR #2341, opened 2019-01-14
remove unused builcraft pipe interaction

temp1011:
It looks like the code hasn't been touched since it was commented out in 1.9, even when builcraft support was added back for most other things. I'm not even sure it is necessary given that engines seem to work nicely with things like itemducts.

Seems like the simplest thing to do is to just remove the code to prevent any more crashes from stripped methods.


AlexIIL:
Just FYI pipes aren't `IItemHandler`'s, so this would still be necessary if you wanted to insert items into BC pipes.

(However as the code does nothing commenting it all out or just commenting out the ``@Optional`` annotation might be a good idea. Basically either merge this or #2344).

temp1011:
The updated api  name fix is better I think.

PR #2463, opened 2019-09-09
[1.12] allow farms to pick up fruit/nuts dropped from decaying leaves

temp1011:
After the changes to leaf drop chance many trees will now drop lots of fruit when grown in a farm. This PR enables the farm to pick up the fruit. The implementation isn't the cleanest but I'm not sure of a better way to do this in the current version.

### DIFF
--- a/src/main/java/forestry/farming/logic/farmables/FarmableGE.java
+++ b/src/main/java/forestry/farming/logic/farmables/FarmableGE.java
@@ -14,6 +14,8 @@ import com.google.common.collect.Sets;
 
 import javax.annotation.Nullable;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import net.minecraft.block.Block;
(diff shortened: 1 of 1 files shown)


src/main/java/forestry/farming/logic/farmables/FarmableGE.java
```diff
@@ -22,17 +28,35 @@
...
 import forestry.arboriculture.ModuleArboriculture;
+import forestry.core.utils.datastructures.ItemStackMap;
+import forestry.core.utils.datastructures.StackMap;
 import forestry.farming.logic.crops.CropDestroy;
 
 public class FarmableGE implements IFarmable {
 
+	//StackMap used because a normal HashSet didn't seem to work
+	//TODO use items instead and normal set if this is still used in 1.14+ since items are flattened
+	private final StackMap<ItemStack, Boolean> windfall = new ItemStackMap<>();
```
mezz: Try using `Collections.newSetFromMap(new ItemStackMap<>())`

mezz (changes requested):
Mostly looks good, you can create a set from a map in Java which should simplify it a little.
The reason ItemStack does not work in a normal Set or Map is because it does not properly implement a hash code or equality check.

src/main/java/forestry/farming/logic/farmables/FarmableGE.java
temp1011: Thanks, I didn't know that existed.

mezz (approved):

PR #2469, opened 2019-09-10
Fix up a couple of methods in genetics

temp1011:
Split out from #2468 since it doesn't explicitly belong there


Nedelosk:
Included in https://github.com/ForestryMC/ForestryMC/pull/2468

Issue #2245, opened 2018-09-08
Fermenter cycle value config partially nonfunctional


lOmicronl:
In the difficulty config, under gamemode -> fermenter -> value, there are two lines that define how much biomass relates to one catalyst cycle (compost/mulch or fertilizer). They default to 48 and 56, respectively.

When I change the value for fertilizer, it works as expected ingame.

When I change the value for compost, it does not actually result in any changes ingame. The fermenter still processes items in chunks of 48 biomass at a time when filled with compost, no matter which value I provided in the config.

lOmicronl:
"enhancement"? I would think that if something was broken, it would at least qualify as a minor bug... =P

temp1011:
Seems that it either only occurs at certain values or requires a couple of restarts to get going from my testing.

lOmicronl:
How odd. When I tested, I changed the values for both fertilizer and compost at the same time, and the fertilizer value took while the compost value did not.

Could it have something to do with repeatedly changing difficulty modes through multiple restarts?

PR #2470, opened 2019-09-10
[1.14] Start generating recipes in data run

temp1011:
This is just some initial work on some of the more annoying recipes to handle in json. Eventually it would probably be nice to have everything generated.

I'm not sure of the best way to move the recipes from generated to a directory where they'll be loaded by the game.

With these changes the total size of the generated directory (including tags etc) is about *?* lines of json. (will update this once all recipes are being generated)


src/main/java/forestry/storage/features/BackpackItems.java
```diff
@@ -14,6 +14,7 @@
 public class BackpackItems {
 	private static final IFeatureRegistry REGISTRY = ModFeatureRegistry.get(Constants.MOD_ID).getRegistry(ForestryModuleUids.BACKPACKS);
 
+	//TODO there is a race condition here: If Backpacks loads before apiculture then the group is null.
```
temp1011: I'm not 100% on the cause but there's definitely something breaking due to `ItemGroups` with these recipes.

Nedelosk: This should be fixed now.
https://github.com/ForestryMC/ForestryMC/blob/23b9a3ca8c24ecf6548a5898f0c102c97c802de8/src/main/java/forestry/core/ItemGroupForestry.java#L39-L40

temp1011: The problem still exists. If I put breakpoints on the static initialiser in `ItemGroupForestry` and `BackpackItems` then `BackpackItems` is loaded first. But I'm not really sure how to solve this without making the ItemGroups static initialisers, which means we can't hide items when the module isn't enabled.

Nedelosk: The easiest fix would be to give @FeatureProvider a priority value. And priorities ItemGroupForestry.

Nedelosk: https://github.com/ForestryMC/ForestryMC/blob/f02df303c2bbd66ab13ef8c212747a845c423755/src/main/java/forestry/modules/CommonModuleHandler.java#L99-L100

temp1011: It works perfectly now, thanks.

src/main/java/forestry/Forestry.java
```diff
@@ -199,12 +201,16 @@ public void setupClient(FMLClientSetupEvent event) {
 	}
 
 	private void gatherData(GatherDataEvent event) {
+		CapabilityFluidHandler.register();	//TODO test
```
temp1011, line 226: Forge calls this event in preInit which is either too late or not called on the data run. So to be able to create recipes with containers I added a call here. Probably forge should fix this.

src/main/java/forestry/core/data/ForestryRecipeProvider.java
```diff
@@ -0,0 +1,1050 @@
...
+			return constructor.newInstance(stack);
+
+		} catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+
+	}
+
+
+	private void registerFluidsRecipes(RecipeDataHelper helper) {
```
temp1011, line 1058: The constructor of `IngredientNBT` isn't public so I have to use reflection to get to it. Again this seems like a forge issue.

temp1011 (commented):
Note of hacks so far

temp1011:
@Nedelosk is it acceptable to make the recipes that depend on the gamemode constant output? Or do we want to keep that feature?

Nedelosk:
I have no idea if anyone uses the game mode feature anymore.

temp1011:
I have no idea either.

temp1011:
I guess this issue mentions them: https://github.com/ForestryMC/ForestryMC/issues/2245

temp1011:
Screenshot from 2019-09-26 23-52-11
Putting this here in case it turns out to be an issue

Nedelosk:
I pulled this into the 1.15.2 dev branch.

thedarkcolour:
Did someone just interact with the CLA assisstant?

PR #2471, opened 2019-09-10
[1.14] Flattening

temp1011:
Mostly cleanup of broken code through porting. Farm block textures/models may be more broken now but shouldn't be too bad to fix.


src/main/java/forestry/farming/ModuleFarming.java
```diff
@@ -242,28 +242,28 @@ public void doInit() {
 
 		Circuits.farmArborealManaged = new CircuitFarmLogic("managedArboreal", arborealFarm, false);
 		Circuits.farmArborealManual = new CircuitFarmLogic("manualArboreal", arborealFarm, true);
-		arborealFarm.registerSoil(new ItemStack(Blocks.DIRT), coreBlocks.humus.getDefaultState());
-		arborealFarm.registerSoil(new ItemStack(coreBlocks.humus), coreBlocks.humus.getDefaultState());
+//		arborealFarm.registerSoil(Blocks.DIRT);	//TODO check this doesn't allow players to use dirt as soil in managed farm
```
temp1011: I'm not really sure how this worked before. Perhaps I can't simplify the API here

Nedelosk: I think it makes more sense to just keep `registerSoil(ItemStack, BlockState)` so other mods could register some sort of item and place a block that is completely unconnected to the item.

temp1011: Ok, I've kept the old API

temp1011:
Blocked on resolution of https://github.com/MinecraftForge/MinecraftForge/issues/6177 so that the minecarts are fixed.

Issue #2318, opened 2018-12-16
Vanilla fix | Potential problem


olesiek:
Hi, I'm suspecting Forestry for getting java.langIO.... bla bla bla exception when trying to join to my minecraft server, Vanilla fix report:

Save  New  Duplicate & Edit  Just Text  Twitter
---- Minecraft Crash Report ----
// You're mean.

Time: 2018-12-16 19:13:28 CET
Description: Error executing task

java.util.concurrent.ExecutionException: java.lang.NullPointerException: Tried to send data to client before the player exists.
    at java.util.concurrent.FutureTask.report(Unknown Source)
    at java.util.concurrent.FutureTask.get(Unknown Source)
    at net.minecraft.util.Util.runTask(SourceFile:531)
    at net.minecraft.client.Minecraft.runGameLoop(Minecraft.java:1087)
    at net.minecraft.client.Minecraft.run(Minecraft.java:3941)
    at net.minecraft.client.main.Main.main(SourceFile:123)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
    at java.lang.reflect.Method.invoke(Unknown Source)
    at net.minecraft.launchwrapper.Launch.launch(Launch.java:135)
    at net.minecraft.launchwrapper.Launch.main(Launch.java:28)
Caused by: java.lang.NullPointerException: Tried to send data to client before the player exists.
    at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:787)
    at forestry.core.network.PacketHandler.lambda$checkThreadAndEnqueue$0(PacketHandler.java:76)
    at forestry.core.network.PacketHandler$$Lambda$4241/1612487367.run(Unknown Source)
    at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
    at java.util.concurrent.FutureTask.run(Unknown Source)
    at net.minecraft.util.Util.runTask(SourceFile:529)
    ... 9 more

A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- System Details --
  Minecraft Version: 1.12.2
  Operating System: Windows 8.1 (amd64) version 6.3
  Java Version: 1.8.0_45, Oracle Corporation
[...]

Issue #1534, opened 2016-12-13
[1.10.2-5.2.16.331] Intermittently forgetting all discovered bees/mutations/research.


ErdrickXLII:
I seem to remember this happening before, but I'm not sure if it was ever resolved.
Sometimes when I start playing Minecraft, my Apiarist's Chest will say I have found zero species/queens/drones/etc.  And sometimes it will load my discovered species and such just fine, but then forget about them later.

Looking into the ApiaristTracker.[blahblahblah].dat file shows that it's still properly keeping track of all discovered and researched mutations, it's just the Apiarist's Chest and Analyzer aren't showing them.

PR #2473, opened 2019-09-18
try and debug network race condition some more, and prevent login crash

temp1011:
cc #2318 
Handling the packet outside the minecraft thread but relying on `Minecraft.player` to be set seems like a race condition, but I don't have a better solution right now. There are two other places using `PlayerLoggedInEvent` but run their code synchronously and things work fine for them.

I believe the race condition occurs when syncing breeding trackers, so perhaps sync lazily when required.

For now I think the best approach is to at least allow players to log in.

### DIFF
--- a/src/main/java/forestry/core/network/PacketHandler.java
+++ b/src/main/java/forestry/core/network/PacketHandler.java
@@ -73,8 +73,12 @@ public class PacketHandler {
 			threadListener.addScheduledTask(() -> {
 				try {
 					EntityPlayer player = Minecraft.getMinecraft().player;
-					Preconditions.checkNotNull(player, "Tried to send data to client before the player exists.");
-					packet.onPacketData(data, player);
+					if(player != null) {
+						packet.onPacketData(data, player);
+					} else {
+						Log.error("Client player was null when packet received: {}, {}", packet.getClass().getCanonicalName(), data.toString());
+						Log.error("Breeding trackers etc will likely be broken in this world");
+					}
 					data.release();
 				} catch (IOException e) {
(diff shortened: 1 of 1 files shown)


mezz:
`threadListener.addScheduledTask` puts this code on the main minecraft thread, so there should not be a race condition there.

I would ask the people reporting to provide a crash log to see if they are still having the same issue, or see if it's something new.

temp1011:
From further investigation I suspect this is caused by vanilla fix. I guess the only question is whether we keep the null check for defensive purposes.

mezz:
If we have null there, things are super extremely broken... unfortunately there isn't much we can do at that point so I think it makes sense to just crash so that we don't cause even weirder and more complex behavior.
Please make sure to report it to vanilla fix, they are responsive to bug reports like this and should be able to help.

temp1011:
Talk about good timing...

I was reading through the forge discord last night and saw this.
chat
So it looks like there is perhaps a vanilla/forge issue here and vanilla fix just forces it to a hard crash (or Barteks2x is wrong).

I think if this is a vanilla issue it would also be the cause of https://github.com/ForestryMC/ForestryMC/issues/1534

So I don't know what to do here, it seems like vanilla fix could actually help by forcing something that would become weird behaviour into a hard crash.

Either way, I have opened a vanilla fix issue [here](https://github.com/DimensionalDevelopment/VanillaFix/issues/210) to try and resolve things

PS - I can't work out how to copy links to messages in the browser version of discord, sorry for the image!

temp1011:
Looks like this isn't going to get resolved any time soon so I will close this.

PR #2483, opened 2019-09-28
[1.12] fix arboretums getting saplings mixed up

temp1011:
I belive this is the simplest solution that makes sure farm logic is reused. I haven't tested with the multifarm yet though, but for single block farms this seems to work.

### DIFF
--- a/src/main/java/forestry/api/farming/IFarmHousing.java
+++ b/src/main/java/forestry/api/farming/IFarmHousing.java
@@ -5,6 +5,7 @@
  ******************************************************************************/
 package forestry.api.farming;
 
+import net.minecraft.item.ItemStack;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Vec3i;
 import net.minecraft.world.World;
@@ -76,4 +77,8 @@ public interface IFarmHousing extends IErrorLogicSource {
 
 	/* GUI */
 	int getStoredFertilizerScaled(int scale);
+
+	default void addPendingProduce(ItemStack stack) {
+
+	}
 }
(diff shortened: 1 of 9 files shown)


Nedelosk:
This would be a fast fix, but I don' think it is a good one. To reduce the number of logic instances was a great idea by mezz. To change it the way you have done it would probably just create new bugs. The right question, in this case, is why is there the need to save the list for one cycle and return it in the next?

temp1011:
I think `produce` is where the sand is added when humus decays so we still need it (moving over to the PR since it's probably the better place)

temp1011:
https://github.com/ForestryMC/ForestryMC/blob/68c52f899f5872d1e56777109afddcd5a130826f/src/main/java/forestry/farming/logic/FarmLogicHomogeneous.java#L68-L106

Nedelosk:
There surely is a bug with` lastExtentsHarvest` too, but it gets never triggered because no one would place two farms so close together.

temp1011:
I've added a fix for the produce possibly not always acting correctly (two sets of drops replacing each other). It's untested though. So I think there is potentially a bug with the produce getting sand from different farms too although I'm not certain.

temp1011:
So I think we have:
- `produce` bug 1: farms set it to each others windfall pickup
- `lastExtents` bug (possibly hard to occur)
- `produce` bug 2: farms set each others drops to sand (potentially, probably hard to occur similar to above)

Nedelosk:
How should this PR fix the bug? Ther are still only two instances of every logic that are saved in the circuits.

temp1011:
It doesn't fix it for circuits, I forgot about those. It does fix it for things that get their logic out of the `FarmRegistry` by `ForestryFarmIdentifier`. I'm not sure how to fix circuits yet. 

I don't mind removing the shared state instead, but I have to work out how to call `maintainSoil` in `FarmLogicHomogeneous` to fix the soil of farms.

Nedelosk:
What about adding a method that adds an `ItemStack` to the `pendingProduce` collection ?

temp1011:
@Nedelosk do I remove all the `lastExtents` as well (quite a few between different logics). I'm not convinced they will be useful if they get polluted by multiple farms.

PR #2503, opened 2019-12-14
Feature/shiftlag

richardhendricks:
Fixes shift-lag for Forestry worktable.
Supports options for
- halting crafting after so many ms
- creating a local cache of recipes
- updating the global recipe map to move recipes to the front that are used, making them faster for everyone.

This may cause problems with recipes that are wildcarded, it is still under evaluation.


richardhendricks:
Sorry, wrong repo

ramidzkh:
bruh its closed

PR #2520, opened 2020-02-07
EscritoireGameToken performance fix

radioegor146:
If the item of EscritoireGameToken is not visible then there is no need to send it


Nedelosk:
I overlooked that this fix is for the version 1.7.10 of forestry. We are no longer maintaining this version. I accidentally accepted this Pr but it will probably not even by build into a version that gets uploaded to curse forge.

Issue #2541, opened 2020-04-11
Humus not turning into sand when a tree grows on it


LuigiHutch:
Forestry_1.12.2-5.8.2.422
Forge-1.12.2-14.23.5.2852

Humus doesnt seem to 'degrade' when a sapling grows on it so it doesnt turn into sand no matter the config

MauveCloud:
I'm curious how you tested this, because I was able to get humus to degrade using Forestry build 422 and Forge build 2854:
![2020-06-15_20 18 14](https://user-images.githubusercontent.com/8942955/84728241-a8557b00-af45-11ea-9523-62c68fcde5fa.png)

However, by my calculations, the median time should be about 40 minutes to degrade one step, though there is a 2.78% chance of it taking more than 3.47 hours.  Also, if you have humus directly below an isolated tree like that, afaik it will never degrade, even with the PR auto-linked above.

PR #2550, opened 2020-05-27
[1.12] Fix for #2541 - humus never turning to sand

tstites14:
This should allow humus to turn into sand again. In the current implementation, each time the humus block is ticked it returns immediately because there is rng that checks if said random number is zero. This effectively means that each tick only has a 1/141 chance of even checking if it is enrooted and thus advancing the degradation. Let me know if there is anything I overlooked or broke in this pr!

### DIFF
--- a/src/main/java/forestry/core/blocks/BlockHumus.java
+++ b/src/main/java/forestry/core/blocks/BlockHumus.java
@@ -79,7 +79,7 @@ public class BlockHumus extends Block implements IItemModelRegister {
 
 	@Override
 	public void updateTick(World world, BlockPos pos, IBlockState state, Random rand) {
-		if (world.isRemote || world.rand.nextInt(140) != 0) {
+		if (world.isRemote || world.rand.nextInt(140) == 0) {
 			return;
 		}
 


Nedelosk:
This is the same way we used to it in the 1.7.10 version. So theoretically there should be another problem if it worked in 1.7.10 and nothing about the random tick code changed massively. 
https://github.com/ForestryMC/ForestryMC/blob/e83dcbc61d22535c5717c97fb7e8161719c37713/src/main/java/forestry/core/blocks/BlockSoil.java#L134-L138

tstites14:
From what I can tell, the random numbers generated when the block ticks are certainly non-zero. I am not sure what would have changed that made them not work, but I am not sure what else it could have been. Here is a quick log of the numbers that I was getting just now from the rng.

```
[16:08:05] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 74
[16:08:11] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 49
[16:08:20] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 55
[16:09:18] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 34
[16:09:23] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 116
[16:10:45] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 20
[16:11:40] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 41
[16:11:59] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 61
[16:12:03] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 40
[16:12:28] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 105
[16:14:17] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 126
[16:14:50] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 122
[16:15:43] [Server thread/INFO] [STDOUT]: [forestry.core.blocks.BlockHumus:updateTick:83]: Random value: 30

```

Nedelosk:
I think the best way of handling it would be to use a new float value and add it to the config. What do you think?

tstites14:
What would this float value be used for or replace? I am not entirely sure what you are suggesting. Could you please clarify?

MauveCloud:
> This effectively means that each tick only has a 1/141 chance of even checking if it is enrooted and thus advancing the degradation.

I'm guessing that was deliberate, to keep the degradation from happening too fast, and maybe also to reduce lag.  What sense does it make to reverse it and only have a 1/140 chance of skipping said check?

I also notice some other differences between the 1.7.10 and 1.12.2 functionality, though I'm not sure whether they're significant:
1. In 1.7.10 the 1/140 random chance was in degradeSoil, now it's in updateTick.
2. The blocks compared are different - 1.7.10 compared against Blocks.log, Blocks.sapling, and IGrowable, so Forestry logs wouldn't match, but 1.12.2 checks against BlockLog and IGrowable.
3. In 1.7.10 it would return false on a match directly above the block before checking the remaining blocks, while in 1.12.2 it skips checking for logs/saplings directly above the block itself.  In both cases, it seems like the decay is dependent on logs or saplings adjacent to directly above it, and an isolated tree would never trigger decay on the humus below it regardless of that world.rand.nextInt call.

Issue #2237, opened 2018-08-30
Request: Config option to disable beehives giving off light


Crowfooted:
See title. Beehives are pretty but I'd like to be able to disable their glowing for atmospheric reasons.

PR #2551, opened 2020-05-28
[1.12] Add config option to modify beehive light level - Issue #2237

tstites14:
Added an option to common.cfg to change the amount of light beehives produce in a range from 0.0f (off) to 0.4f (the original value).


Nedelosk:
Thank you for your help. But wouldn't it make sense to allow numbers higher than 0.4 too?

tstites14:
That's fair. I was not sure what value should be the cap so I simply put it as the value it was before. What number do you suggest it should be so I can change it?

Nedelosk:
Why not 1.0F ? And could you please add a config option comment ("for.config.performance.hiveLightLevels.comment") to the lang file ?

PR #2564, opened 2020-07-26
Updated ru_ru.lang

Radviger:
Updated ru_ru.lang
Fixed research note chat message translation
Fixed duplicated (client+server) multiblock error message


Radviger:
Has some translation errors.

PR #2619, opened 2021-11-30
Fix woven backpack recipes and make them use tags

ramidzkh:
Makes the backpacks use tags for accept/reject and also fix the woven backpack recipes


PickleOnAString:
nice

good job

ramidzkh:
Merged with d6f6b72d50c92d0188f705b0fea8ed5503c9e3f1

PR #2646, opened 2022-03-09
Fix Patchouli and optimize getting recipes

mezz:
Another PR, build on top of the JEI one: https://github.com/ForestryMC/ForestryMC/pull/2645

This one fixes some crashes in Patchouli so that the book sorta works again.

### DIFF
--- a/src/main/java/forestry/core/patchouli/processor/CarpenterProcessor.java
+++ b/src/main/java/forestry/core/patchouli/processor/CarpenterProcessor.java
@@ -1,58 +1,34 @@
 package forestry.core.patchouli.processor;
 
-import java.util.ArrayList;
-import java.util.List;
-
-import net.minecraft.world.item.ItemStack;
-import net.minecraft.world.item.crafting.Ingredient;
-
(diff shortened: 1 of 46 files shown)


src/main/java/forestry/core/patchouli/processor/CarpenterProcessor.java
```diff
@@ -1,58 +1,34 @@
...
-		try {
-			itemStack = variables.get("item").as(ItemStack.class);
-		} catch (Exception e) {
-			itemStack = ItemStack.EMPTY;
-		}
-
-        /*
-        Manually iterate over all the carpenter recipes by result item ResourceLocation id
-        because looking up the recipe in carpenterManager doesn't work for some recipes
-        (idk why, maybe i was doing it wrong)
```
mezz, line 38: @ramidzkh you weren't doing anything wrong here, the existing method just had a really strict equality check for ItemStack. I removed that existing method, because I think ItemStack comparison should always be left up to the caller, since every situation is different.

PR #2651, opened 2022-04-10
Fixes for the forestry manual

thehutch:
Fix several issues with the manual where there were missing images. I made these fixes a while ago, but it does appear that the book code has been deleted? Will it be re-added at some future point? If so, the data fixes will still be useful


ramidzkh:
The manual code was removed in favor of using Patchouli

PR #2736, opened 2025-02-06
Buff to the bee tick rate and production method for bees' products

dreais:
The problems and possible solutions I tried to implement:
- Bee tick rate was set to 550t/bee cycle (27.5s). This is way too high to properly scale this, adding to that the 1 per cycle *at maximum* for the specialty output (and 2 for primary output)
    - Solution: set the default to 40. This *should* be possible to change it in the config file, I removed the hardcoded value just to be safe, I honestly have no idea if this was the issue.
- Products output was capped, 1 per primary, 1 per secondary, 1 per specialty. This is especially bad since you already have a very long waiting time for ONE output, but even stacking more productivity and speed does nothing to it
    - Solution: changed the formula to have a productivity based drop instead of a chance based drop. Meaning that a chance of 350% chance/productivity will yield 3 products + 50% of a 4th.
    - The productivity formula is based off of GT:NH [(here)](https://gtnh.miraheze.org/wiki/Bees), but removed the tier based formula
    - I did end up removing the Secondary output, but not using pure-breed feels more punishing and random, which just seems against scaling up

Essentially, the idea of this as a whole is to scale up the production rather than the time factor. For Gregtech players especially, this makes it a competitive option for ore generation, and/or a nice complement. 

I did add a cap on how many frames an alveary can have, at 4, as the lower tick rate already makes it worth just running 1 alveary and adding more if necessary. For mods integrating with Forestry such as Greg, better frames also scale higher, which is a fine compromise I think

Here are some numbers:

Frame(s) | Product Chance (%) | Housing Productivity | Bee Speed | Output (cycle)
-- | -- | -- | -- | --
No Frame | 30% | 100% | 60% | 0,3686585536
1x 200% Frame | 30% | 200% | 60% | 0,4645113526
2x 200% Frame | 30% | 400% | 60% | 0,6215128608
3x 200% Frame | 30% | 800% | 60% | 0,8688712912
4x 200% Frame | 30% | 1600% | 60% | 1,248513097
4x 400% Frame (Gregtech) | 30% | 25600% | 60% | 5,870824496
4x 1000% Frame (Gregtech) | 30% | 1000000% | 60% | 47,32187793

### DIFF
--- a/src/main/java/forestry/api/apiculture/IBeeModifier.java
+++ b/src/main/java/forestry/api/apiculture/IBeeModifier.java
@@ -8,6 +8,13 @@ package forestry.api.apiculture;
 import javax.annotation.Nullable;
 
 public interface IBeeModifier {
+
+	enum ProductionType {
+		FRAME,
+		OTHER
+	}
+
+	ProductionType TypeDefault = ProductionType.FRAME;
 	/**
 	 * @param genome          Genome of the bee this modifier is called for.
 	 * @param currentModifier Current modifier.
--- a/src/main/java/forestry/apiculture/ModuleApiculture.java
+++ b/src/main/java/forestry/apiculture/ModuleApiculture.java
@@ -155,7 +155,8 @@ public class ModuleApiculture extends BlankForestryModule {
 
 	public static String beekeepingMode = "NORMAL";
 
-	public static int ticksPerBeeWorkCycle = 550;
+	// Previous value was 550 (~27 IRL seconds) - way too high. It's better IMO to scale down the production (or up) rather than have a very long tick rate and an enormous output at once
+	public static int ticksPerBeeWorkCycle = 40;
 
 	public static boolean hivesDamageOnPeaceful = false;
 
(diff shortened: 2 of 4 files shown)


dreais:
Adding [cc974f4](https://github.com/ForestryMC/ForestryMC/pull/2736/commits/cc974f4afffbac0d6cdb11751a9a18b6ab6f450e)

I initially had a hard limit on how many frames could be accepted but I realized it would also count towards electrons. I just added a quick and dirty enum to possibly implement in binnie's bees (or other) to bypass that as they can be multiplicative to each others without impacting the total productivity

src/main/java/forestry/apiculture/ModuleApiculture.java
```diff
@@ -155,7 +155,9 @@ public class ModuleApiculture extends BlankForestryModule {
 
 	public static String beekeepingMode = "NORMAL";
 
-	public static int ticksPerBeeWorkCycle = 550;
+	// Previous value was 550 (~27 IRL seconds) - way too high. It's better IMO to scale down the production (or up) rather than have a very long tick rate and an enormous output at once
+	// Is set later from the config file, by default is set to 40 ticks
+	public static int ticksPerBeeWorkCycle;
```
thedarkcolour, line 160: The default should be left at 550. A config option would be a better alternative, so that GregTech players can scale the work cycle ticks without affecting the rest of the playerbase. Keep in mind that reducing this number reduces the in-game lifespan of a bee. I also don't think the upper limit should be reduced.

src/main/java/forestry/api/apiculture/IBeeModifier.java
```diff
@@ -8,6 +8,13 @@
 import javax.annotation.Nullable;
 
 public interface IBeeModifier {
+
+	enum ProductionType {
+		FRAME,
+		OTHER
```
thedarkcolour, line 14: Could you explain a bit more about the purpose of OTHER?

